PR #62, opened 2024-10-03
feat(document-search): Allow to create DocumentSearch instances from config

akotyla:


packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -58,6 +62,29 @@ def __init__(
...
+            DocumentSearch: An initialized instance of the DocumentSearch class.
+        """
+
+        embedder = get_embeddings(config["embedder"])
+        query_rephraser = get_rephraser(config.get("rephraser"))
+        reranker = get_reranker(config.get("reranker"))
+        vector_store = get_vector_store(config["vector_store"])
+
+        providers_config_dict: dict = config.get("providers", {})
+        providers_config = DocumentProcessorRouter.from_dict_to_providers_config(providers_config_dict)
```
mhordynski, line 83: That looks off to me, probably should be more something like `ProvidersConfig.from_dict()` ?

akotyla: Can it be simply extracted to a separate function outside of any class? because now we don't have a `ProvidersConfig` class, it's just a type `dict[DocumentType, BaseProvider]` and if we created such a class, it would have only one attribute - this dictionary, and methods, apart from that `.from_config` would be `get` and `update` to handle this dictionary and it seems to me that it's maybe a bit of an unnecessary abstraction, what do you think?

mhordynski (approved):
lgtm

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mhordynski: ahh I see, let's leave it like this for now

PR #74, opened 2024-10-08
feat(document-search): allow to use local instance of unstructured

konrad-czarnota-ds:
I added the local unstructured as default (I think running samples for users shouldn't force them to provide api key or server endpoints which may be seen as a potential blocker). To run with api one needs to create `DocumentSearch` with argument: `document_processor_router=DocumentProcessorRouter.from_config({DocumentType.TXT: UnstructuredProvider(use_api=True)})`


mhordynski (approved):
lgtm

PR #78, opened 2024-10-09
chore: Add package version update action

akotyla:
Additionally, the script for updating packages versions has been fixed, so it also work for the `ragbits` package.


mhordynski (approved):
lgtm

PR #87, opened 2024-10-11
chore: use ruff in the project

PatrykWyzgowski:


PatrykWyzgowski:
Please mind that [too-many-newlines-at-end-of-file (W391)](https://docs.astral.sh/ruff/rules/too-many-newlines-at-end-of-file/#too-many-newlines-at-end-of-file-w391) is still in preview mode.

packages/ragbits-core/src/ragbits/core/prompt/promptfoo.py
```diff
@@ -21,6 +21,9 @@ def generate_configs(
         root_path: The root path to search for Prompt objects. Defaults to the directory where the script is run.
         target_path: The path to save the promptfoo configuration files. Defaults to "promptfooconfigs".
     """
+    if not root_path:
+        root_path = Path.cwd()
```
akonarski-ds, line 25: Can we be consistent with this?  Now we have three ways of doing it, above we use `root_path = root_path or Path.cwd()`, below we have a seprate func for it (`_get_current_dir`). Let's choose one and stick to it

akonarski-ds (approved):

packages/ragbits-core/src/ragbits/core/llms/clients/local.py
```diff
@@ -74,10 +72,9 @@ async def call(
         conversation: ChatFormat,
         options: LocalLLMOptions,
         json_mode: bool = False,
-        output_schema: Optional[Type[BaseModel] | Dict] = None,
+        output_schema: type[BaseModel] | dict | None = None,
     ) -> str:
-        """
-        Makes a call to the local LLM with the provided prompt and options.
+        """Makes a call to the local LLM with the provided prompt and options.
```
micpst, line 77: Why are these docstrings formatted this way? In some places they are formatted with a new line like this:
```
"""
some docstring
"""
```
I think we should stick to one format (preferably the latter).

packages/ragbits-core/src/ragbits/core/prompt/parsers.py
```diff
@@ -87,7 +88,9 @@ def bool_parser(value: str) -> bool:
     raise ResponseParsingError(f"Could not parse '{value}' as a boolean")
 
 
-def build_pydantic_parser(model: Type[PydanticModelT]) -> Callable[[str], PydanticModelT]:
+def build_pydantic_parser(
```
micpst, line 91: Why is this line reformatted?

packages/ragbits-core/tests/unit/prompts/test_prompt.py
```diff
@@ -124,7 +124,10 @@ class TestPrompt(Prompt[_PromptInput, str]):  # pylint: disable=unused-variable
     assert prompt.rendered_system_prompt == "You are a song generator for a adult named Alice."
     assert prompt.rendered_user_prompt == "Theme for the song is rock."
     assert prompt.chat == [
-        {"role": "system", "content": "You are a song generator for a adult named Alice."},
+        {
```
micpst, line 127: Again some formatting issues...

micpst:
In general, most of the code should remain unchanged, please fix the formatting problems.

packages/ragbits-core/src/ragbits/core/llms/clients/base.py
```diff
@@ -12,9 +12,9 @@
 
 
 @dataclass
-class LLMOptions(ABC):
```
micpst, line 15: Why is it no longer an abstract class?

micpst: Matter of preference, here we simply don't want the user to create an instance of `LLMOptions`, cuz we only support specific implementations for implemented clients.

ludwiktrammer: I agree with @micpst - for my marking a class as abstract is not only about being able to mark methods as abstract, but even more importantly about communicating that the class is not intended to be used directly

packages/ragbits-document-search/src/ragbits/document_search/ingestion/document_processor.py
```diff
@@ -49,7 +50,7 @@ def from_config(cls, providers_config: Optional[ProvidersConfig] = None) -> "Doc
         {
             DocumentType.TXT: YourCustomProviderClass(),
             DocumentType.PDF: UnstructuredProvider(),
-        }
+        }.
```
ludwiktrammer, line 53: This seems unusual - dot added to a line of example code. Maybe the code should be in a code block to avoid this?

packages/ragbits-core/src/ragbits/core/llms/clients/base.py
akonarski-ds: Ok, that makes sense. Initially, in this example, I didn't realize that this class isn't meant to be used directly. I thought it could be used in baseline/no-options clients. So, I assume we're expecting that all clients will have some configuration options, and even if not, the user should subclass this class, even without adding options.

.pre-commit-config.yaml
```diff
@@ -26,7 +26,8 @@ repos:
         # Formats Python, Pyi, and Jupyter files, excluding specified directories
         - id: ruff-format
           types_or: [ python, pyi, jupyter ]
-          exclude: (docs/)
+          exclude: (docs/|/tests/)
```
akonarski-ds, line 29: Why? I don't see any reason why tests shouldn't be correctly formatted.

PatrykWyzgowski: Fixed.

ludwiktrammer (approved):

PR #99, opened 2024-10-14
feat(document-search): determine document type automatically

akotyla:


packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -89,7 +90,15 @@ async def ingest_document(
             document_processor: The document processor to use. If not provided, the document processor will be
                 determined based on the document metadata.
         """
-        document_meta = document if isinstance(document, DocumentMeta) else document.metadata
+
+        if isinstance(document, Source):
+            local_path = await document.fetch()
+            document_meta = DocumentMeta.from_local_path(local_path)
```
mhordynski, line 96: hmm, it's not going to work like that. 

When ingesting document from a given source (for example `gcs`) we should have that information persisted along with DocumentMeta. `DocumentMeta.from_local_path` creates new `LocalFileSource` so it loses that important piece of info

akotyla: ok, good point, changed

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
```diff
@@ -97,6 +97,24 @@ def from_local_path(cls, local_path: Path) -> "DocumentMeta":
             source=LocalFileSource(path=local_path),
         )
 
+    @classmethod
+    async def from_source(cls, source: Union[LocalFileSource, GCSSource]) -> "DocumentMeta":
```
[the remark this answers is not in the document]
akotyla: We agreed that this will be fixed in a separate task: https://github.com/deepsense-ai/ragbits/issues/108

mhordynski (approved):
lgtm

PR #106, opened 2024-10-16
feat(sources): add hf data source

micpst:


packages/ragbits-document-search/src/ragbits/document_search/documents/sources.py
```diff
@@ -86,6 +89,7 @@ def get_id(self) -> str:
         """
         return f"gcs:gs://{self.bucket}/{self.object_name}"
 
+    @requires_dependencies(["gcloud.aio.storage"], "gcs")
```
micpst, line 92: I am testing a new way of handling optional dependencies. Since managing error messages from extra deps is starting to get tedious, I decided to wrap them in `@requires_dependencies` decorator that handles them automatically. 

If you like it, we can add a new issue to refactor the handling of optional deps, there are several other places where this decorator should be applied.

packages/ragbits-document-search/src/ragbits/document_search/documents/sources.py
```diff
@@ -99,26 +103,80 @@ async def fetch(self) -> Path:
...
+                file.write(data["content"])  # type: ignore
+
+        return path
+
+
+def get_local_storage_dir() -> Path:
+    """
+    Get the local storage directory.
+
+    The local storage directory is determined by the environment variable `LOCAL_STORAGE_DIR_ENV`. If this environment
```
akotyla, line 172: above you changed the name of this variable to `LOCAL_STORAGE_DIR`

micpst: fixed, thx for catching

packages/ragbits-document-search/src/ragbits/document_search/documents/sources.py
```diff
@@ -86,39 +89,94 @@ def get_id(self) -> str:
...
+        storage_dir = get_local_storage_dir()
+        source_dir = storage_dir / Path(data["source"]).parent  # type: ignore
+        source_dir.mkdir(parents=True, exist_ok=True)
+        path = storage_dir / data["source"]  # type: ignore
+
+        if not path.is_file():
+            with open(path, mode="w", encoding="utf-8") as file:
+                file.write(data["content"])  # type: ignore
+
+        return path
```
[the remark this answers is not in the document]
micpst: `datasets` caches all downloaded datasets out of the box. 

Edit, I changed the implementation slightly and now we use streaming, so we only download the row that is requested by the user.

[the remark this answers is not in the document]
akonarski-ds: I didn't mean to cache `datasets` as I know it downloads and keeps data locally as a cache. What I meant is that every time we call `fetch` we **load** the entire dataset (now it's only one row) into memory.

But after changes it's fine.

packages/ragbits-core/src/ragbits/core/utils/decorators.py
```diff
@@ -0,0 +1,55 @@
...
+import asyncio
+from functools import wraps
+from importlib.util import find_spec
+from typing import Callable, ParamSpec, TypeVar
+
+_P = ParamSpec("_P")
+_T = TypeVar("_T")
+
+
+def requires_dependencies(
```
akonarski-ds, line 12: I'd add some tests for this func. It should be fairly easy: we can pass `pytest` or `ragbits` as deps as they are always present, and some made up deps such as `micpst_super_secret_lib` to raise `ImportError`

akonarski-ds (approved):
lgtm

packages/ragbits-core/src/ragbits/core/utils/decorators.py
micpst: done, thx

PR #109, opened 2024-10-16
feat(document-search): LiteLLM Reranker

PatrykWyzgowski:


packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/base.py
```diff
@@ -1,21 +1,23 @@
 import abc
 
+from pydantic import BaseModel
+
 from ragbits.document_search.documents.element import Element
 
 
-class Reranker(abc.ABC):
+class Reranker(BaseModel, abc.ABC):
```
akotyla, line 8: maybe a stupid question, but why is `pydantic` needed here?

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/litellm.py
```diff
@@ -0,0 +1,51 @@
...
+
+        Raises:
+            ValueError: If chunks are not a list of TextElement objects.
+        """
+        if not all(isinstance(chunk, TextElement) for chunk in chunks):
+            raise ValueError("All chunks must be TextElement instances")
+
+        documents = [chunk.content if isinstance(chunk, TextElement) else None for chunk in chunks]
+
+        response = litellm.rerank(
```
akotyla, line 39: maybe it would be better to use `arerank` here: https://docs.litellm.ai/docs/rerank#async-usage, what do you think?

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/litellm.py
```diff
@@ -0,0 +1,51 @@
...
+
+import litellm
+
+from ragbits.document_search.documents.element import Element, TextElement
+from ragbits.document_search.retrieval.rerankers.base import Reranker
+
+
+class LiteLLMReranker(Reranker):
+    """
+    A LiteLLM reranker for providers such as Cohere, Together AI, Azure AI.
```
akotyla, line 11: I think it might be helpful to describe a bit these attributes below or link the documentation, I couldn't find `litellm`, but `cohere` is probably the same https://docs.cohere.com/v2/reference/rerank#request

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/litellm.py
```diff
@@ -0,0 +1,51 @@
+from typing import List
+
+import litellm
+
+from ragbits.document_search.documents.element import Element, TextElement
+from ragbits.document_search.retrieval.rerankers.base import Reranker
+
+
+class LiteLLMReranker(Reranker):
```
akotyla, line 9: you should add it here: https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/__init__.py#L9 for it to work with config like this:
```
"reranker": {
        "type": "LiteLLMReranker",
        "config": {"model": "cohere/rerank-english-v3.0"},
    }
```

packages/ragbits-document-search/examples/reranker_example.py
micpst, line 1: Maybe let's not introduce another example here, just put the reranker configuration in `examples/document-search/from-config.py`.

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/litellm.py
```diff
@@ -0,0 +1,63 @@
...
+        return_documents: Whether to return the documents or just indices and relevalnce scores.
+        rank_fields: The fields to rank on.
+        max_chunks_per_doc: The maximum number of chunks per document to consider.
+
+    Returns:
+        The reranked chunks.
+    """
+
+    model: str
+    top_n: int | None = None
```
micpst, line 27: These params should be in a separate data class (look at the `LLMOptions` implementation)

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/litellm.py
```diff
@@ -0,0 +1,63 @@
+from typing import List
+
+import litellm
+from pydantic import BaseModel
+
+from ragbits.document_search.documents.element import Element, TextElement
+from ragbits.document_search.retrieval.rerankers.base import Reranker
+
+
+class LiteLLMReranker(BaseModel, Reranker):
```
micpst, line 10: Reranker shouldn't be a dataclass

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/litellm.py
```diff
@@ -0,0 +1,63 @@
...
+            chunks: The chunks to rerank.
+            query: The query to rerank the chunks against.
+
+        Returns:
+            The reranked chunks.
+
+        Raises:
+            ValueError: If chunks are not a list of TextElement objects.
+        """
+        if not all(isinstance(chunk, TextElement) for chunk in chunks):
```
micpst, line 46: Why only `TextElements` here?

mhordynski: I think that is partially because of that how `Element` base model is designed. What we're lacking is some sort of abstract method that would return text representation of the element. (e.g. `.to_text()`). Then we could use that unified interface while passing elements to LLMs, rerankers, etc.

mhordynski: Created #143

micpst (approved):

PR #112, opened 2024-10-16
feat(document-search): batch ingestion from sources

ludwiktrammer:


packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -128,7 +128,44 @@ async def ingest_document(
         if document_processor is None:
             document_processor = self.document_processor_router.get_provider(document_meta)
 
-        elements = await document_processor.process(document_meta)
+        document_processor = self.document_processor_router.get_provider(document_meta)
+        return await document_processor.process(document_meta)
+
+    async def ingest_document(
```
ludwiktrammer, line 134: I would probably want to remove `ingest_document` all together. And then maybe even rename `ingest_documents` to simply `ingest`. What do you think?

micpst: I like the idea of a single endpoint + `ingest` is concise, I like it.

packages/ragbits-document-search/src/ragbits/document_search/documents/sources.py
```diff
@@ -66,6 +66,23 @@ async def fetch(self) -> Path:
...
+            path: The path to the directory.
+            file_pattern: The file pattern to match.
+
+        Returns:
+            List of source objects.
+        """
+        sources = []
+        for file_path in path.glob(file_pattern):
+            sources.append(cls(path=file_path))
+        return sources
```
akonarski-ds, line 84:
```suggestion
        return [cls(path=file_path) for file_path in path.glob(file_pattern)]
```

packages/ragbits-document-search/tests/unit/test_local_file_source.py
```diff
@@ -0,0 +1,21 @@
...
+
+from ragbits.document_search.documents.sources import LocalFileSource
+
+TEST_FILE_PATH = Path(__file__)
+
+
+async def test_local_source_fetch():
+    source = LocalFileSource(path=TEST_FILE_PATH)
+
+    path = await source.fetch()
```
akonarski-ds, line 11:
```suggestion
    path = await source.fetch()
    
```
to follow the AAA convention

packages/ragbits-document-search/tests/unit/test_local_file_source.py
```diff
@@ -0,0 +1,21 @@
...
+    assert path == TEST_FILE_PATH
+
+
+async def test_local_source_list_sources():
+    example_files = TEST_FILE_PATH.parent / "example_files"
+    sources = LocalFileSource.list_sources(example_files, file_pattern="*.md")
```
akonarski-ds, line 17:
```suggestion

    sources = LocalFileSource.list_sources(example_files, file_pattern="*.md")
```
to follow the AAA convention

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -128,7 +128,27 @@ async def ingest_document(
...
+        """
+        Ingest multiple documents.
+
+        Args:
+            documents: The documents or metadata of the documents to ingest.
+            document_processor: The document processor to use. If not provided, the document processor will be
+                determined based on the document metadata.
+        """
+
+        elements = []
```
micpst, line 148:
```suggestion
        elements = await asyncio.gather(*[self._process_document(document, document_processor) for document in documents])
```

mhordynski (approved):
lgtm

packages/ragbits-document-search/src/ragbits/document_search/_main.py
ludwiktrammer: I thought about it but opted for adding TODO instead, because each `self._process_document` may (and probably will) lead to network requests and uncontrolled paralyzation of a big number of network requests would not be a good idea and may even lead to failed requests.

micpst: Ok, makes sense

micpst (approved):

PR #120, opened 2024-10-20
refactor(cli): lazy-load modules in cli on command level

mhordynski:


packages/ragbits-core/src/ragbits/core/cli.py
```diff
@@ -1,21 +1,52 @@
...
+        target_path: Path = Path("promptfooconfigs"),
+    ) -> None:
+        """
+        Generates the configuration files for the PromptFoo prompts.
+
+        This function registers all the CLI modules in the ragbits packages:
+            - iterates over every package in the ragbits.* namespace
+            - it looks for `cli` package / module
+            - if found it imports the `register` function from the `cli` module and calls it with the `app` object
+            - register function should add the CLI commands to the `app` object
```
akonarski-ds, line 46: It seems like a docstring from a different function

akonarski-ds (approved):
apart from the comment about docstring, lgtm

packages/ragbits-core/src/ragbits/core/cli.py
mhordynski: thx fixed

ludwiktrammer (approved):

PR #121, opened 2024-10-21
feat(document-search): add support for images

konrad-czarnota-ds:


packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/default.py
```diff
@@ -132,20 +137,26 @@ async def process(self, document_meta: DocumentMeta) -> list[Element]:
...
-    return env_value
+        return await self._chunk_and_convert(elements, document_meta, document.local_path)
+
+    async def _chunk_and_convert(
+        self, elements: list[UnstructuredElement], document_meta: DocumentMeta, document_path: Path
+    ) -> list[Element]:
+        if self.__class__ == UnstructuredDefaultProvider:
+            chunked_elements = chunk_elements(elements, **self.chunking_kwargs)
+            return [to_text_element(element, document_meta) for element in chunked_elements]
+        image_elements = [e for e in elements if e.category == "Image"]
```
ludwiktrammer, line 148: You should probably use Unstructured's `ElementType.Image`  instead of a raw string

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/default.py
```diff
@@ -132,20 +137,26 @@ async def process(self, document_meta: DocumentMeta) -> list[Element]:
...
-        return value
-    if (env_value := os.getenv(env_var)) is None:
-        raise ValueError(f"Either pass {name} argument or set the {env_var} environment variable")
-    return env_value
+        return await self._chunk_and_convert(elements, document_meta, document.local_path)
+
+    async def _chunk_and_convert(
+        self, elements: list[UnstructuredElement], document_meta: DocumentMeta, document_path: Path
+    ) -> list[Element]:
+        if self.__class__ == UnstructuredDefaultProvider:
```
ludwiktrammer, line 145: This looks confusing to me. Is this about having a behavior on the base class that the classes inheriting from it don't share? 

Inheriting a class and having the behavior change magically, even though I haven't changed it myself, would be surprising and not something I would like to encounter as a programmer. Seems like this behavior should be in a method that inheriting class can overwrite if they want.

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/default.py
```diff
@@ -70,13 +71,16 @@ def __init__(
                 variable will be used.
             api_server: The API server URL to use for the Unstructured API. If not specified, the
                 UNSTRUCTURED_SERVER_URL environment variable will be used.
+            use_api: whether to use unstructured API
```
ludwiktrammer, line 74: A more detailed documentation would be helpful - what is the alternative that will be used when `use_api` is set to `False`?

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/pdf.py
```diff
@@ -0,0 +1,82 @@
...
+        """
+        super().__init__(partition_kwargs, chunking_kwargs, api_key, api_server, use_api)
+        self.image_summarizer = ImageDescriber(
+            llm_model_name or DEFAULT_LLM_IMAGE_SUMMARIZATION_MODEL, llm_options or DEFAULT_LLM_OPTIONS
+        )
+
+    async def _to_image_element(
+        self, element: UnstructuredElement, document_meta: DocumentMeta, document_path: Path
+    ) -> ImageElement:
+        image_coordinates = extract_image_coordinates(element)
```
ludwiktrammer, line 62: nit: unpacking in a way that the coordinates have meaningful names might make the code clearer (when I see `image_coordinates[0]` I'm not sure what I'm looking at)

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    Returns:
+        text element
+    """
+    return TextElement(
+        document_meta=document_meta,
+        content=element.text,
+    )
+
+
+def set_or_raise(name: str, value: Optional[str], env_var: str) -> str:
```
ludwiktrammer, line 31: The name, the description, and the actual behavior don't much, at least when it comes to my mental model.

- I would expect a function named `set_or_raise` to set a value and if it can't - to raise an exception. I also wouldn't know that it has anything to do with environmental variables
- I would expect a function described as "Checks if given environment variable is set and returns it or raises an error" to simply look for an env variable and if it's not present - to raise an exception
- In reality this is more like a fallback mechanism - first it tries to use the value given as an argument, if it's not present it tries to get it from the environment, and if it's not present there it throws an exception. The argument called "name" is not actually used anywhere other than the error message, and the error message makes it clear that the function is about checking arguments. I would have never gotten all this without reading the code. 

Maybe rename to something like `check_required_argument(value: Optional[str], arg_name: str, fallback_env: str)`

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/utils.py
```diff
@@ -0,0 +1,124 @@
...
+class ImageDescriber:
+    """
+    Describes images content using an LLM
+    """
+
+    DEFAULT_PROMPT = "Describe the content of the image."
+
+    def __init__(self, model_name: str, model_options: LiteLLMOptions):
+        self.model_name = model_name
+        self.model: Optional[LiteLLM] = None
```
ludwiktrammer, line 92: nit: I think we usually refer to `LLM` objects as `llm`, not `model`

ludwiktrammer (changes requested):

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/default.py
konrad-czarnota-ds: Both inheriting classes need to do the same preprocessing:
```
        image_elements = [e for e in elements if e.category == ElementType.IMAGE]
        other_elements = [e for e in elements if e.category != ElementType.IMAGE]
        chunked_other_elements = chunk_elements(other_elements, **self.chunking_kwargs)

        text_elements: list[Element] = [to_text_element(element, document_meta) for element in chunked_other_elements]
```
I didn't want to duplicate this code in both classes nor introduce a third "in-between" class so I came up with this idea. Happy to change it if you have a better suggestion how to handle it.

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/utils.py
konrad-czarnota-ds: This is actually a function that I simply moved from another class, it came without documentation so I wrote something myself, but I guess it wasn't enough. Changed to your suggestion.

packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -1,10 +1,10 @@
 from abc import ABCMeta, abstractmethod
-from typing import Dict, Generic, List, Optional, Type
+from typing import Any, Dict, Generic, List, Optional, Type
 
 from pydantic import BaseModel
 from typing_extensions import TypeVar
 
-ChatFormat = List[Dict[str, str]]
+ChatFormat = List[Dict[str, Any]]
```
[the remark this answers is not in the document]
ludwiktrammer: Thank change it to `list[dict[str, str | list[dict]]]`. `Any` is very rarely the right answer.

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/default.py
ludwiktrammer: I would definitely change this - it's a very surprising behavior and you don't want unexpected surprises from your code. Even though both classes that **currently** inherit from `UnstructuredDefaultProvider` need this behavior, the base class should be ready to be subclassed any time for other reasons. It shouldn't contain code specific to functionality that is not related to its intended behavior and definitely shouldn't magically switch to it for all of its children. 

This is an important principle: child classes are downstream from the base class and they can know and use the specific of the base class. On the other hand, the base class is upstream from its children. It cannot know when it will be subclassed and for what reason. 

Seems that `UnstructuredPdfProvider` and `UnstructuredImageProvider` are similar because they both deal with images - the only difference is where they get the images from.

You can reflect that in the inheritance structure. Maybe pdf provider is just an image provider enhanced with the ability to find images in PDFs? Than the inheritance could look like that:

```
UnstructuredDefaultProvider -> UnstructuredImageProvider -> UnstructuredPdfProvider
``` 

If this doesn't seem correct to you in this case, alternatively you could have an intermediate (possibly abstract) class that adds image-related functionality, inherits from `UnstructuredDefaultProvider` and is a base for both `UnstructuredImageProvider` and `UnstructuredPdfProvider`

packages/ragbits-core/src/ragbits/core/prompt/base.py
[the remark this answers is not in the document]
konrad-czarnota-ds: Unfortunately it didn't help, but introduced one more error from mypy. I tried to fight with it a bit but without success and decided to simply add ignore line there (this is to be changed anyway to use our prompt template with image support)

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/default.py
konrad-czarnota-ds: Thanks, that turned out to be a good idea, now pdf extends the images and there are only two small functions that needed to be implemented

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/images.py
```diff
@@ -0,0 +1,111 @@
...
+from ragbits.document_search.documents.element import Element, ImageElement
+from ragbits.document_search.ingestion.providers.unstructured.default import UnstructuredDefaultProvider
+from ragbits.document_search.ingestion.providers.unstructured.utils import (
+    ImageDescriber,
+    crop_and_convert_to_bytes,
+    extract_image_coordinates,
+    to_text_element,
+)
+
+DEFAULT_LLM_IMAGE_SUMMARIZATION_MODEL = "gpt-4o-mini"
```
ludwiktrammer, line 20: This is not a blocker for this PR but maybe this should use our built-in `core_config.default_llm_factory` functionality? Or if not (because there is no guarantee that `default_llm_factory` returns LLM that supports images) maybe we should have a separate config option for default visual LLM?

konrad-czarnota-ds: Created [an issue](https://github.com/deepsense-ai/ragbits/issues/140) for that, as discussed.

PR #122, opened 2024-10-21
feat(document-search) add location metadata to document

kdziedzic68:


packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
```diff
@@ -50,6 +60,19 @@ def id(self) -> str:
         """
         return self.source.id
 
+    def add_location_metadata(self, provider_metadata: dict | None) -> None:
```
mhordynski, line 63: why `| None` here?

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
```diff
@@ -32,13 +32,23 @@ class DocumentType(str, Enum):
     UNKNOWN = "unknown"
 
 
+class DocumentLocation(BaseModel):
```
mhordynski, line 35: I'm wondering if that model applies for each document type that we support. For example for PDF files location may differ from MD location

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
kdziedzic68: in unstructured output format metadata was marked as optional AFAIR

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
kdziedzic68: the metadata for unstructured output has document agnostic format. have inspected it and did not see anything that could be document type dependent - would leave a comment there to potentially keep in mind for the future

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
```diff
@@ -32,13 +32,23 @@ class DocumentType(str, Enum):
...
+
+
 class DocumentMeta(BaseModel):
     """
     An object representing a document metadata.
     """
 
     document_type: DocumentType
     source: LocalFileSource | GCSSource | HuggingFaceSource = Field(..., discriminator="source_type")
+    location: DocumentLocation | None = None
```
micpst, line 51: I think this should be defined in the `Element` class

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
micpst: I don't see optional typing for `element.metadata` anywhere, but still you use this method with `model_dump` in the condition, so you're sure it's a dict here

packages/ragbits-document-search/src/ragbits/document_search/documents/document.py
kdziedzic68: done

micpst (approved):

PR #146, opened 2024-10-24
feat(core): add CLI commands to render / exec prompts 

akotyla:


packages/ragbits-core/src/ragbits/core/cli.py
```diff
@@ -43,4 +61,34 @@ def generate_promptfoo_configs(
...
+        """
+        Renders a prompt by loading a class from a module and initializing it with a given payload.
+        """
+        prompt = _render(prompt_path=prompt_path, payload=payload)
+
+        pprint("[orange3]RENDERED PROMPT:")
+        pprint(prompt.chat)
+
+    @prompts_app.command()
+    def execute(
```
akotyla, line 75: I called it `execute` instead of `exec`, because with `exec` I got `redefined-builtin` from `pylint` and I don't know if we want that

mhordynski: You can use `@prompts_app.command(name="exec")` to override function name

akotyla: thanks, changed

mhordynski (approved):
lgtm

PR #148, opened 2024-10-24
refactor: Add dynamic loading for modules that depend on optional dependencies

### LINKED ISSUE #110: refactor: Do not import optional dependencies in __init__.py
We shouldn't import code that depends on optional dependencies, as we do in:

- [ingestion providers](https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/__init__.py)
- [embeddings](https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-core/src/ragbits/core/embeddings/__init__.py)
- [llms](https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-core/src/ragbits/core/llms/__init__.py)
- [vector stores](https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-core/src/ragbits/core/vector_store/__init__.py)

because it has performance implications due to many unnecessary imports

akonarski-ds:
Closes #110

### DIFF
--- a/docs/api_reference/core/embeddings.md
+++ b/docs/api_reference/core/embeddings.md
@@ -2,6 +2,6 @@
 
 ::: ragbits.core.embeddings.Embeddings
 
-::: ragbits.core.embeddings.LocalEmbeddings
+::: ragbits.core.embeddings.local.LocalEmbeddings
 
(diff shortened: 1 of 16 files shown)


mhordynski (approved):
lgtm

PR #149, opened 2024-10-24
feat: Prompt and LLM should accept images

kdziedzic68:


packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -1,10 +1,10 @@
 from abc import ABCMeta, abstractmethod
-from typing import Dict, Generic, Optional, Type
+from typing import Any, Dict, Generic, Optional, Type
 
 from pydantic import BaseModel
 from typing_extensions import TypeVar
 
-ChatFormat = list[dict[str, str]]
+ChatFormat = list[dict[str, Any]]
```
kdziedzic68, line 7: adding images to the prompt changes the format: if the images are present than content of message is a list of dictionaries, was trying to annotate it more precisely, however went through terrible things with mypy and done this change to speed things up. if precise annotations are important - let me know, however would prefer not to invest time into such things

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -26,6 +26,7 @@ def __init__(self, model_name: str, default_options: Optional[LLMOptions] = None
             TypeError: If the subclass is missing the '_options_cls' attribute.
         """
         self.model_name = model_name
+        self.has_vision = False
```
ludwiktrammer, line 29: I don't think this is necessary - each LLM client implementation can decide whether (for any reason) it cannot handle a given request and react accordingly in this `call` method implementation (similarly how they handle other problems there already).

If we ever want `LLM` objects to report their capabilities, we should do it in a less ad-hoc manner, so it's not vision-specific. But I believe that can wait.

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -66,8 +67,12 @@ async def generate_raw(
 
         Returns:
             Raw text response from LLM.
+        Raises:
+            ValueError: prompt containing images and model does not support vision
```
ludwiktrammer, line 71: As I mentioned during daily, I imagined this differently (handling the request sans the images, with a warning) but this is also okay. I have no problem with throwing an exception. But then we should create a more specific exception that is specifically for cases where the model doesn't have required capabilities, so that the code that calls the client can easily detect and handle such cases.

packages/ragbits-core/src/ragbits/core/prompt/base.py
ludwiktrammer: I think this is pretty important because this is a part of the public API of our library - we use `ChatFormat` in multiple places to document the required format for arguments / the return format of various methods.

packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -37,6 +37,15 @@ def output_schema(self) -> Optional[Dict | Type[BaseModel]]:
         """
         return None
 
+    def list_images(self) -> list[dict]:
```
ludwiktrammer, line 40: I believe this should simply return a list of images, without the return format being coupled with the complicated LLM conversation format.

This is part of our public API and a method that our users will be expected to implement whenever they implement custom prompt sources. Is should be as simple as possible.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -23,6 +24,7 @@ class Prompt(Generic[InputT, OutputT], BasePromptWithParser[OutputT], metaclass=
 
     system_prompt: Optional[str] = None
     user_prompt: str
+    images: Optional[List[bytes]] = None
```
ludwiktrammer, line 27: I don't think this would be useful - here you define images statistically, which means that they will be the same for every instance of the prompt. I don't believe this solves the usual use case - usually images are part of the _input data_, not the static prompt definition.

That's why the ticket says:

> The Prompt class should include its own implementation of the method, that automatically returns images from chosen fields of the **Prompt's input model** (which fields should probably be configured as a static field in the Prompt class)

The aim is to extract the images from the input data, not having them defined statically as part of the class definition.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -181,6 +183,44 @@ def list_few_shots(self) -> ChatFormat:
...
+        if not images:
+            user_message = {
+                "role": "user",
+                "content": self.rendered_user_prompt,
+            }
+        else:
+            user_message = {
+                "role": "user",
+                "content": [{"type": "text", "text": self.rendered_user_prompt}, *images],  # type: ignore
+            }  # type: ignore
```
ludwiktrammer, line 221: Why is `# type: ignore` necessary here?

ludwiktrammer (changes requested):
Could you also [resolve the TODO commit here](https://github.com/deepsense-ai/ragbits/blob/d79ef6d866695c7a57b9b67ba85e22fc34213b5d/packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/utils.py#L105) by changing the code to use your new `Prompt` implementation? It should be easy and will also serve as a test that the implementation serves the desired use-case.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -181,6 +183,44 @@ def list_few_shots(self) -> ChatFormat:
...
+                    "image_url": {"url": f"data:image/png;base64 {base64.b64encode(image).decode('utf-8')}"},
+                }
+                for image in self.images
+            ]
+            if self.images
+            else []
+        )
+        return image_prompt_list
+
+    def get_user_message(self) -> ChatFormat:
```
ludwiktrammer, line 205: Despite the name, this is not a getter for the user message, so this may be confusing to our users. Either way, I would make this method private - it seems to be an internal method only meant to be used by the `chat` method

packages/ragbits-core/src/ragbits/core/prompt/base.py
ludwiktrammer: To be clear: I mean it as "it's work giving a good try", not "this is a must and we should spend an infinite amount of time on it" :)

kdziedzic68: this has been resolved in the latest version

packages/ragbits-core/src/ragbits/core/llms/base.py
kdziedzic68: already outdated - not used anymore

packages/ragbits-core/src/ragbits/core/llms/base.py
kdziedzic68: warnings are raised on level of llm client now

packages/ragbits-core/src/ragbits/core/prompt/base.py
kdziedzic68: done

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
kdziedzic68: this has been removed from here however `# type: ignore` is used at the other part of the code - the reason is that dict representing image has different struture than representing text - mypy keeps throwing errors if not used.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
kdziedzic68: has been already removed - processing of image input has been moved to level of specific llm client

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -119,14 +130,17 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
...
+        self.images = self._get_images_from_input_data(input_data)
 
         # Additional few shot examples that can be added dynamically using methods
         # (in opposite to the static `few_shots` attribute which is defined in the class)
         self._instace_few_shots: list[FewShotExample[InputT, OutputT]] = []
         super().__init__()
 
     @property
-    def chat(self) -> ChatFormat:
+    def chat(self) -> ChatFormat | ChatFormatWithImages:
```
ludwiktrammer, line 143: Why? Is the only reason the fact that I insisted that having a precise typing for `ChatFormat` would be a good thing, or is there another reason?

Having a method that _sometimes_ returns a data structure that is ready to be used with LLMs, and _sometimes_ returns an anonymous tuple that needs to be processed (and will fail when given directly to the LLM) makes the API much, much more confusing. If I were to choose between an imprecise type annotation and this, I would definitely choose an imprecise type annotation.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -74,6 +75,16 @@ def _render_template(cls, template: Template, input_data: Optional[InputT]) -> s
             context = input_data.model_dump(serialize_as_any=True)
         return template.render(**context)
 
+    def _get_images_from_input_data(self, input_data: Optional[InputT]) -> list:
+        images = []
+        if isinstance(input_data, BaseModel):
+            context = input_data.model_dump(serialize_as_any=True)
```
ludwiktrammer, line 81: Why do you serialize the input data model? There is no need for this - you can inspect the object directly, without loosing the precise typing and metadata, and without spending time on serialization.

packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -1,21 +1,47 @@
...
 
 from pydantic import BaseModel
 from typing_extensions import TypeVar
 
 ChatFormat = list[dict[str, str]]
+ChatFormatWithImages = tuple[ChatFormat, list[bytes]]
 OutputT = TypeVar("OutputT", default=str)
 
 
+class InputType(BaseModel):
```
ludwiktrammer, line 12: I don't understand why this Pydantic model is needed (and why it is public). Seems that that it holds static data (and not user-provided data that needs to be validated and/or serialized) that could be simple a static variable in this file.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -119,14 +130,17 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             self._render_template(self.system_prompt_template, input_data) if self.system_prompt_template else None
         )
         self.rendered_user_prompt = self._render_template(self.user_prompt_template, input_data)
+        self._image_input_type = InputType(collection_type=list, element_type=bytes)
```
ludwiktrammer, line 133: I wonder whether type `bytes` is enough to know that the field contains an image. For example, in the future if we allow attaching audio to prompts, most likely the audio data will also be provided as `bytes`.

I would either:
* Create [a custom Pydantic data type](https://docs.pydantic.dev/2.3/usage/types/custom/) for images, that would allow us to detect them unequivocally
* Leave automatic detection of images for now, and allow users to provide `image_fields` by hand

ludwiktrammer (changes requested):

ludwiktrammer:
> Could you also [resolve the TODO commit here](https://github.com/deepsense-ai/ragbits/blob/d79ef6d866695c7a57b9b67ba85e22fc34213b5d/packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/utils.py#L105) by changing the code to use your new `Prompt` implementation? It should be easy and will also serve as a test that the implementation serves the desired use-case.

Just to make sure - have you seen this comment above?

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
kdziedzic68: `ChatFormatWithImages` has been removed - even two lines of # type: ignore were enough to overcome issue with definition of `ChatFormat`

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
kdziedzic68: has been addressed

packages/ragbits-core/src/ragbits/core/prompt/base.py
kdziedzic68: removed

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
kdziedzic68: done - currently we support definition of `image_fields` with class variables

packages/ragbits-core/src/ragbits/core/llms/clients/base.py
```diff
@@ -4,7 +4,7 @@
 
 from pydantic import BaseModel
 
-from ragbits.core.prompt import ChatFormat
+from ragbits.core.prompt.base import ChatFormat
```
ludwiktrammer, line 7: Why?

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -83,3 +85,24 @@ def count_tokens(self, prompt: BasePrompt) -> int:
...
+        images = prompt.list_images()
+        chat = prompt.chat
+        if images:
+            if not litellm.supports_vision(self.model_name):
+                warnings.warn(
+                    message=f"Model {self.model_name} does not support vision. Image input would be ignored",
+                    category=UserWarning,
+                )
+                return chat
+            user_message = chat[-1]
```
ludwiktrammer, line 99: We cannot safely assume that the last message is the `user_message`. There may be few show examples. You should actually look at the type specified in the message.

packages/ragbits-core/src/ragbits/core/prompt/base.py
ludwiktrammer: As of right now it seems to be still here

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -83,3 +85,24 @@ def count_tokens(self, prompt: BasePrompt) -> int:
             Number of tokens in the prompt.
         """
         return sum(litellm.token_counter(model=self.model_name, text=message["content"]) for message in prompt.chat)
+
+    def _format_chat_for_llm(self, prompt: BasePrompt) -> ChatFormat:
```
ludwiktrammer, line 89: `ChatFormat` defined as `list[dict[str, str]]` is not the correct type here - this causes you to have to sprinkle `# type: ignore` throughout the documentation, but more importantly is not a true representation of what this method returns. Either use a separate type defined as `list[dict[str, Any]]` here, or change the `ChatFormat` definition.

packages/ragbits-core/src/ragbits/core/llms/litellm.py
kdziedzic68: actually in prompt.chat we always add user message at the end - few shots (if exist) are added before

packages/ragbits-core/src/ragbits/core/llms/clients/base.py
kdziedzic68: removed

packages/ragbits-core/src/ragbits/core/llms/litellm.py
ludwiktrammer: Ok, I still don't feel that blindly taking the last message is reasonable, when they have their types specified explicitly. This is a big assumption that may be correct today, but doesn't have to true moving forward - since the types are specified, let's make use of the fact.

packages/ragbits-core/src/ragbits/core/llms/litellm.py
kdziedzic68: changed

packages/ragbits-core/src/ragbits/core/prompt/base.py
kdziedzic68: oops. now its removed

packages/ragbits-core/src/ragbits/core/llms/litellm.py
kdziedzic68: ok - so the goal of code below is to add image input to the user message. should we than:

a) search for the last user message in the conversation and add image input to it ?
b) if the last message type is not user ignore it and raise some warning ?

kdziedzic68: btw - regarding few shots i think we should create a backlog ticket to support images in few shots as well

ludwiktrammer: > btw - regarding few shots i think we should create a backlog ticket to support images in few shots as well

Definitely - I wanted to suggest the same 😃

ludwiktrammer: Another related feature for a separate ticket is allowing `Prompts` with no text user-message in cases when there are images specified (i.e., support for scenario when the prompt consists of system prompt and image-only user prompt)

packages/ragbits-core/src/ragbits/core/llms/clients/base.py
ludwiktrammer: Seems unchanged

packages/ragbits-core/src/ragbits/core/llms/clients/local.py
```diff
@@ -11,7 +11,7 @@
 except ImportError:
     HAS_LOCAL_LLM = False
 
-from ragbits.core.prompt import ChatFormat
+from ragbits.core.prompt.base import ChatFormat
```
ludwiktrammer, line 14: Here as well

packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/unstructured/utils.py
```diff
@@ -85,35 +86,18 @@ class ImageDescriber:
...
-                        "type": "image_url",
-                        "image_url": {"url": f"data:image/jpeg;base64,{img_base64}"},
-                    },
-                ],
-            }
-        ]
-        return await self.llm.client.call(messages, self.llm.default_options)  # type: ignore
+        if not prompt.list_images():
+            wrngs.warn(message="Image data not provided", category=UserWarning)
+        return await self.llm.generate(prompt=prompt)  # type: ignore
```
ludwiktrammer, line 103: Seems to me this ` # type: ignore`  is no longer necessary, isn't it?

packages/ragbits-core/src/ragbits/core/llms/litellm.py
ludwiktrammer: @kdziedzic68 both (a) and (b) are ok with me - so whatever you prefer

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -74,6 +75,16 @@ def _render_template(cls, template: Template, input_data: Optional[InputT]) -> s
...
         return template.render(**context)
 
+    @classmethod
+    def _get_images_from_input_data(cls, input_data: Optional[InputT]) -> list[bytes]:
+        images = []
+        if isinstance(input_data, BaseModel):
+            image_input_fields = cls.image_input_fields or []
+            for field in image_input_fields:
+                if field in input_data.model_fields:
+                    images.extend(getattr(input_data, field))
```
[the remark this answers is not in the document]
kdziedzic68: done

packages/ragbits-core/src/ragbits/core/llms/clients/local.py
kdziedzic68: fixed

packages/ragbits-core/src/ragbits/core/llms/clients/base.py
kdziedzic68: fixed

ludwiktrammer (approved):

PR #151, opened 2024-10-27
refactor: chroma client

micpst:


packages/ragbits-core/src/ragbits/core/vector_store/chromadb_store.py
```diff
@@ -61,114 +70,62 @@ def from_config(cls, config: dict) -> ChromaDBStore:
...
-                )
-
-                db_entries.append(db_entry)
+        options = self._default_options if options is None else options
+        results = self._collection.query(
+            query_embeddings=vector,
+            n_results=options.k,
+            include=["metadatas", "embeddings", "distances"],
+        )
+        metadatas = results.get("metadatas") or []
```
akonarski-ds, line 117:
```suggestion
        metadatas = results.get("metadatas", [])
```

akonarski-ds (approved):
lgtm

packages/ragbits-core/src/ragbits/core/vector_store/chromadb_store.py
micpst: I was thinking about this, but I couldn't infer proper types with this, cuz `get` always returns `Optional`

akonarski-ds: ok, so let's leave it as it is

PR #153, opened 2024-10-28
feat(core): user should be able to configure default types of default…

konrad-czarnota-ds:
… LLMs

### DIFF
--- a/packages/ragbits-core/tests/unit/llms/factory/test_get_default_llm.py
+++ b/packages/ragbits-core/tests/unit/llms/factory/test_get_default_llm.py
@@ -1,6 +1,7 @@
 import pytest
 
 from ragbits.core.config import core_config
+from ragbits.core.llms.base import LLMType
 from ragbits.core.llms.factory import get_default_llm
 from ragbits.core.llms.litellm import LiteLLM
 
@@ -10,7 +11,7 @@ def test_get_default_llm(monkeypatch: pytest.MonkeyPatch) -> None:
     Test the get_llm_from_factory function.
     """
     monkeypatch.setattr(
-        core_config, "default_llm_factories", {"text": "factory.test_get_llm_from_factory.mock_llm_factory"}
+        core_config, "default_llm_factories", {LLMType.TEXT: "factory.test_get_llm_from_factory.mock_llm_factory"}
     )
 
     llm = get_default_llm()
--- a/packages/ragbits-core/tests/unit/llms/factory/test_get_default_llm.py
+++ b/packages/ragbits-core/tests/unit/llms/factory/test_get_default_llm.py
@@ -9,7 +9,9 @@ def test_get_default_llm(monkeypatch: pytest.MonkeyPatch) -> None:
     """
     Test the get_llm_from_factory function.
     """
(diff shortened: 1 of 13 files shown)


packages/ragbits-core/tests/unit/llms/factory/test_get_default_llm.py
```diff
@@ -1,6 +1,7 @@
 import pytest
```
ludwiktrammer, line 2: I think it would be worth adding couple of tests - to check that getting different types of LLMs work (and not only the default text type) and that using an nonexistent type in pyproject generates an exception

pyproject.toml
```diff
@@ -183,3 +183,8 @@ known-third-party = [
     "requests", "scipy", "setuptools", "shapely", "skimage", "sklearn", "streamlit",
     "torch", "torchvision", "tqdm", "typer"
 ]
+
+[tool.ragbits.core.default_llm_factories]
```
ludwiktrammer, line 187: Why do we want this in the library's main pyproject?

packages/ragbits-core/tests/unit/llms/factory/test_get_default_llm.py
konrad-czarnota-ds: Tests added

pyproject.toml
konrad-czarnota-ds: Without this tests are failing, some of them require this default functions + I think its good to have an example how to set it up for the users

packages/ragbits-core/tests/unit/utils/pyproject/testprojects/factory_project/pyproject.toml
```diff
@@ -0,0 +1,7 @@
+[project]
+name = "factory_project"
+
+[tool.ragbits.core.default_llm_factories]
+"ragbits.core.llms.base.LLMType.TEXT" = "ragbits.core.llms.factory.simple_litellm_factory"
```
ludwiktrammer, line 5: Sorry that I haven't caught this in the previous round of reviews, but could the format be the same as in the ticket, i.e. by enum values:

```
[tool.ragbits.core.default_llm_factories]
text = "ragbits.core.llms.factory.simple_litellm_factory"
vision = "ragbits.core.llms.factory.simple_litellm_vision_factory"
structured_outputs = "ragbits.core.llms.factory.simple_litellm_vision_factory"
```

Having to specify full path for each key seems unnecessary and complicated for the users. We already know that each key is a value in `ragbits.core.llms.base.LLMType`.

pyproject.toml
ludwiktrammer: Which tests? This seems like a problem with the tests - it means that the test will give different results depending the directory from which the command is run, which is not good.

ludwiktrammer (changes requested):

pyproject.toml
ludwiktrammer: Ok. This should test should fail, because it alerts us to a bug in the implementation.

You changed the previous implementation to rely on default image LLM provider to always be set, which is not a true assumption for user projects. We actually want tests that makes sure that the library doesn't fail for projects without default LLM providers specified in their pyproject.toml files.

ludwiktrammer (approved):

PR #156, opened 2024-10-28
refactor(document-search): change the type in from_source method to Source

ludwiktrammer:


konrad-czarnota-ds (approved):
LGTM

packages/ragbits-document-search/src/ragbits/document_search/documents/sources.py
```diff
@@ -43,13 +63,48 @@ async def fetch(self) -> Path:
             The path to the source.
         """
 
+    @classmethod
+    def __init_subclass__(cls, **kwargs: Any) -> None:  # noqa: ANN401
+        Source._registry[cls.class_identifier()] = cls
+        super().__init_subclass__(**kwargs)
+
+
+@dataclass
```
micpst, line 72: Why dataclass here?

ludwiktrammer: Good catch - it was a case of mindlessly coping the documentation. I started by copying an example of custom annotation form Pydantic documentation and then changed the logic. The example was marked a dataclass because it had a configurable field, which I removed, but forgot to remove the dataclass decorator.

ludwiktrammer: Removed

micpst (approved):

PR #164, opened 2024-10-31
chore: add pre-commit/push hooks install scripts

mhordynski:


PatrykWyzgowski (approved):

CONTRIBUTING.md
```diff
@@ -31,18 +31,12 @@ $ uv run mypy .
 ```
 
 
-## Install pre-commit
+## Install pre-commit or pre-push hooks
```
akonarski-ds, line 34: What about other checks e.g. if large file is added to a repository:
https://github.com/deepsense-ai/ragbits/blob/main/.pre-commit-config.yaml#L12
Can we somehow install them here too?

scripts/install_git_hooks.py
```diff
@@ -0,0 +1,76 @@
...
+#!/usr/bin/env bash
+
+echo "🧹 Running formatting...\n"
+uv run ruff format --check
+
+if [ $? -ne 0 ]
+then
+    echo "⚠ Formatting failed. Running autofix & aborting..."
+    uv run ruff format
+    exit 1
```
akonarski-ds, line 27: IMO, it would be better to exit at the end, after all of the checks have been run

mhordynski: Initially I had it like this, but mypy feels really slow - so I decided that it's better experience to fail fast in case of format / linting issues

akonarski-ds: ok, makes sense

CONTRIBUTING.md
mhordynski: We have a couple options here:

1. Invoke both `pre-commit` framework and our custom hook - I think pre-commit tries to do this when installed **after** our hook - but it fails:  
```bash
⟩ pre-commit install

Running in migration mode with existing hooks at .git/hooks/pre-commit.legacy
Use -f to use only pre-commit.
pre-commit installed at .git/hooks/pre-commit
```

```bash
⟩ .git/hooks/pre-commit

### error information

An unexpected error has occurred: OSError: [Errno 8] Exec format error: '/home/mateusz/workspace/oss/ragbits/.git/hooks/pre-commit.legacy'
```

```
# log file

Traceback (most recent call last):
  File "/home/mateusz/workspace/oss/ragbits/.venv/lib/python3.10/site-packages/pre_commit/error_handler.py", line 73, in error_handler
    yield
  File "/home/mateusz/workspace/oss/ragbits/.venv/lib/python3.10/site-packages/pre_commit/main.py", line 389, in main
    return hook_impl(
  File "/home/mateusz/workspace/oss/ragbits/.venv/lib/python3.10/site-packages/pre_commit/commands/hook_impl.py", line 265, in hook_impl
    retv, stdin = _run_legacy(hook_type, hook_dir, args)
  File "/home/mateusz/workspace/oss/ragbits/.venv/lib/python3.10/site-packages/pre_commit/commands/hook_impl.py", line 43, in _run_legacy
    return subprocess.run(cmd, input=stdin).returncode, stdin
  File "/home/mateusz/.local/share/uv/python/cpython-3.10.15-linux-x86_64-gnu/lib/python3.10/subprocess.py", line 503, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/home/mateusz/.local/share/uv/python/cpython-3.10.15-linux-x86_64-gnu/lib/python3.10/subprocess.py", line 971, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/mateusz/.local/share/uv/python/cpython-3.10.15-linux-x86_64-gnu/lib/python3.10/subprocess.py", line 1863, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
OSError: [Errno 8] Exec format error: '/home/mateusz/workspace/oss/ragbits/.git/hooks/pre-commit.legacy'
```

2. Use pre-commit as intented, so figure out how to invoke our hooks from inside pre-commit properly (https://pre-commit.com/#new-hooks) - it is kinda going back, so probably we don't want to do that

3. Install hooks that we care separatly and run them in similiar fashion to ruff and mypy.

akonarski-ds: If we can make option 1 work, that would be great; but if not, I'll go with option 3

Also I don't mind moving it to a separate PR

mhordynski: Yeah, moved here: #166

PR #168, opened 2024-10-31
feat(audit): observability toolset

micpst:


examples/document-search/chroma_otel.py
```diff
@@ -0,0 +1,93 @@
+# /// script
```
mhordynski, line 1: I think we should start adding more comments to the examples. It would be sufficient to write a little summary / docstring on module level explaining what we're doing in the script

micpst: done

packages/ragbits-core/src/ragbits/core/audit/otel.py
```diff
@@ -0,0 +1,95 @@
+from opentelemetry import trace
+from opentelemetry.trace import Span, StatusCode, TracerProvider
+from opentelemetry.util.types import AttributeValue
+
+from ragbits.core.audit.base import TraceHandler
+
+
+class OtelTraceHandler(TraceHandler[Span]):
```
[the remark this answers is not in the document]
micpst: done

mhordynski (approved):
lgtm

PR #169, opened 2024-11-04
enabler: guardrails component design

konrad-czarnota-ds:


mhordynski:
Shouldn't we move guardrails to separate package `ragbits.guardrails`? It seems like standalone code, not necessarily required in `ragbits.core`

micpst (approved):

PR #171, opened 2024-11-05
feat(documents): unified API for text representations of Element models

konrad-czarnota-ds:


packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -36,6 +36,15 @@ def get_key(self) -> str:
             The key.
         """
```
mhordynski, line 37: wdyt about having get_text_representation result as default key?

konrad-czarnota-ds: good idea, changed

mhordynski (approved):
lgtm

Issue #160, opened 2024-10-30
feat(documents): id field for `Element` 


micpst:
### Feature description

Add new `id` computed field for `Element` class. This task also includes switching from sha256 to uuid for id generation.

### Motivation

Currently we compute ids for vector store entries in VectorStore class. This doesn't scale well because we have to re-implement this logic for each implementation, and also introduces some ambiguity.

PR #172, opened 2024-11-05
feat(document-search): Support for ingesting images

ludwiktrammer:


packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -163,5 +164,16 @@ async def insert_elements(self, elements: list[Element]) -> None:
             elements: The list of Elements to insert.
         """
         vectors = await self.embedder.embed_text([element.get_key() for element in elements])
+
+        image_elements = [element for element in elements if isinstance(element, ImageElement)]
```
konrad-czarnota-ds, line 168: What about the previous case support for description of images? In this case we would include image elements twice (one with the upper loop over all elements and second with this loop). Maybe we should add some parameter that would allow for method selection?

uv.lock
```diff
@@ -1,10 +1,18 @@
 version = 1
 requires-python = ">=3.10"
 resolution-markers = [
-    "python_full_version < '3.11'",
-    "python_full_version == '3.11.*'",
-    "python_full_version == '3.12.*'",
-    "python_full_version >= '3.13'",
+    "python_full_version < '3.11' and platform_system == 'Darwin'",
```
konrad-czarnota-ds, line 4: Do we need all those changes in this file?

ludwiktrammer: `uv.lock` is generated automatically by `uv`, it's hard (or impossible) to control it by hand. It seems that some nvidia-related packages introduced some platform-specific versioning

packages/ragbits-document-search/src/ragbits/document_search/_main.py
ludwiktrammer: > What about the previous case support for description of images?

I wanted to preserve the previous method, that's why I don't filter out image elements when creating the `vectors` list. You cannot have one embedding that covers both the text content and the image content, so it seemed reasonable to me to use two vectors to cover both.

I'm open to making this configurable, but maybe this configuration should be on the level of document processor - because running OCR and ignoring it during ingestion would be pretty wasteful

konrad-czarnota-ds: Sounds good to me, we can discuss it with the broader group as well. Feel free to add a separate task for it and merge this one or add it to this already existing PR.

konrad-czarnota-ds (approved):
LGTM

packages/ragbits-core/src/ragbits/core/embeddings/base.py
```diff
@@ -17,3 +17,24 @@ async def embed_text(self, data: list[str]) -> list[list[float]]:
...
         """
+
+    def image_support(self) -> bool:  # noqa: PLR6301
+        """
+        Check if the model supports image embeddings.
+
+        Returns:
+            True if the model supports image embeddings, False otherwise.
+        """
+        return False
```
micpst, line 28: Maybe we can make it dynamic to reduce boilerplate code. I'd check here if `embed_image` is implemented or not and we can set `image_support` as `@property` here.

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
```diff
@@ -0,0 +1,163 @@
...
+    HAS_LITELLM = False
+
+from ragbits.core.embeddings import Embeddings
+from ragbits.core.embeddings.exceptions import (
+    EmbeddingResponseError,
+    EmbeddingStatusError,
+)
+
+
+class VertexAIMultimodelEmbeddings(Embeddings):
```
micpst, line 20: I don't get why we have a separate class for this. We are using litellm under the hood here, it'd be more intuitive to implement `embed_image` in `LiteLLMEmbeddings` and constrain it to VertexAI embeddings.

packages/ragbits-core/src/ragbits/core/embeddings/base.py
ludwiktrammer: The logic here may be a lot more complicated. Specifically in many implementations whether the embedder can support embedding an image will depend on the specific model name and version given. For example my version of `LiteLLMEmbeddings` with image embedding support had this implementation of `image_support`:

```
    def image_support(self) -> bool:
        """
        Check if the model supports image embeddings.

        Returns:
            True if the model supports image embeddings, False otherwise.
        """
        # For now only VertexAI multimodal embeddings are supported for multimodal embeddings
        # We should revisit this once LiteLLM provides a unified API for multimodal embeddings
        return self.model in self.SUPPORTED_MULIMODAL_MODELS
```

(Where `SUPPORTED_MULIMODAL_MODELS` is a static property sourced from LiteLLM's `VertexMultimodalEmbedding().SUPPORTED_MULTIMODAL_EMBEDDING_MODELS`)

If it were this simple - that given class either supports images or not - I wouldn't have `image_support` at all and I wouldn't have added `embed_image` to the base embedding class. Instead, I would define a new abstract interface `EmbeddingsWithImageSupport` (inheriting from the base embeddings class) and added `embed_image` only there. I would then check whether an embedder has image support using `isinstance`. Unfortunately, that's out of the question, because whether a class support image embedding or not may be dynamic and depend on arguments to the constructor.

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
ludwiktrammer: That's the problem I tried to described during the daily today. In theory we still use LiteLLM but with this model LiteLLM behaves differently in almost every way you could imagine:

1. It expects a different input format
2. It returns output in a different format
3. It throws different exceptions when errors occur
4. It doesn't support embedding multiple things at once (causing me to emulate it with async, looping and semaphores in `VertexAIMultimodelEmbeddings`) 

I have a working implementation of `LiteLLMEmbeddings` with VertexAI Multimodal Embeddings support that has workarounds around points 1 and 2. Issue 4 caused me to abandon this direction altogether. Not only having both implementations (the one that can send multiple items to the model at once, and the one that emulates this with async, looping and semaphores) in a  single class would bring a lot of complexity to it, but even more importantly I don't have a condition that would allow me to switch between those two behaviors in a way that makes sense in a general implementation (i.e., other than switching based on the exact name of a single, niche model). That's why, for now, I decided to keep this separate from `LiteLLMEmbeddings`, in order to keep its implementation general and sane. I think we should revisit this in the future, but in my mind having a support for images in `LiteLLMEmbeddings` would require helping LiteLLM deal at least with some of the issues listed above - i.e., helping it provide an unified interface that we could use.

packages/ragbits-document-search/src/ragbits/document_search/_main.py
ludwiktrammer: @konrad-czarnota-ds

I created https://github.com/deepsense-ai/ragbits/issues/175

Looks ok?

konrad-czarnota-ds: Looks good

examples/document-search/multimodal.py
```diff
@@ -0,0 +1,73 @@
+# /// script
```
micpst, line 1: Please add description to this example, should follow the format from [chroma_otel.py](https://github.com/deepsense-ai/ragbits/blob/main/examples/document-search/chroma_otel.py).
> I think we should start adding more comments to the examples. It would be sufficient to write a little summary / docstring on module level explaining what we're doing in the script

_Originally posted by @mhordynski in https://github.com/deepsense-ai/ragbits/pull/168#discussion_r1830614524_

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
```diff
@@ -0,0 +1,163 @@
...
+                input=[instance],
+                model=f"{self.VERTEX_AI_PREFIX}{self.model}",
+                api_base=self.api_base,
+                api_key=self.api_key,
+                **self.options,
+            )
+
+        return response
+
+    async def embed_text(self, data: list[str]) -> list[list[float]]:
```
micpst, line 120: We've merged audit module, this method and `embed_image` should be decorated with `@traceable` or use the `trace()` context manager to log inputs and outputs. 

Ref implementation for [`LiteLLMEmbeddings`](https://github.com/deepsense-ai/ragbits/blob/e674abc8b6837cb2fba3f135c27755c5d7955908/packages/ragbits-core/src/ragbits/core/embeddings/litellm.py#L73-L105)

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
micpst: Ok, makes sense to me

packages/ragbits-core/src/ragbits/core/embeddings/base.py
micpst: Ok, got it

examples/document-search/multimodal.py
ludwiktrammer: Done

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
ludwiktrammer: Does this make sense?

micpst: Looks good, I'd remove the `concurency` because we will see it in the traces

packages/ragbits-core/src/ragbits/core/vector_stores/in_memory.py
```diff
@@ -25,7 +25,7 @@ def __init__(
             metadata_store: The metadata store to use.
         """
         super().__init__(default_options=default_options, metadata_store=metadata_store)
-        self._storage: dict[str, VectorStoreEntry] = {}
+        self._storage: list[VectorStoreEntry] = []
```
micpst, line 28: I'm not sure about this change, with the list we allow duplicates in storage

ludwiktrammer: That's intentional. I see preventing multiple documents with the same textual key as a bug. Textual representations are not unique. We can have multiple different images with the same description (or even: no description, as described in #175). And even for textual documents - the store shouldn't silently loose documents just because they have the same content as other document. They have different metadata and point to different sources. If we wanted to add them all to the store, there was a reason for that, and it's not for the store to decide.

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
ludwiktrammer: Ok, removed

packages/ragbits-core/src/ragbits/core/vector_stores/in_memory.py
micpst: I agree, which is why we have open ticket #160 to address this. Your fix doesn't solve the problem we have with Chroma vector store, where we also can't load duplicate keys due to collisions. 

There is also a risk that a store implemented as a list may grow indefinitely with the same ingest calls (very unlikely, but not impossible). My take is to leave it as it is and solve #160 and we should be good.

ludwiktrammer: Then #160 would become a blocker for merging this PR because the script would become non-functional - only the single last image would ever be stored.

ludwiktrammer: (and it's also unclear to me how generating a random uuid would help with the same objects being ingested multiple time - as far as I understand, each digestion would simply get a new random uuid)

micpst: yeah, uuid doesn't solve the problem here, but changing the components of sha256 should help (add element location data)

PR #180, opened 2024-11-08
docs(document-search): create docs

konrad-czarnota-ds:


docs/how-to/document-search/use_rephraser.md
```diff
@@ -0,0 +1,41 @@
...
+```python
+from ragbits.core.llms.litellm import LiteLLM
+from ragbits.document_search.retrieval.rephrasers.llm import LLMQueryRephraser
+from ragbits.document_search.retrieval.rephrasers.prompts import QueryRephraserPrompt
+
+document_search = DocumentSearch(
+    query_rephraser=LLMQueryRephraser(LiteLLM("gpt-3.5-turbo"), QueryRephraserPrompt),
+    ...
+)
+results = await document_search.search("<query>")
```
kdziedzic68, line 20: maybe we should show how to use rephraser independently too ? if it is not intended to be used directly perhaps we should not include such components in public api

docs/how-to/document-search/use_reranker.md
```diff
@@ -0,0 +1,44 @@
...
+```python
+import os
+from ragbits.document_search.retrieval.rerankers.litellm import LiteLLMReranker
+
+os.environ["COHERE_API_KEY"] = "<api_key>"
+
+document_search = DocumentSearch(
+    reranker=LiteLLMReranker("cohere/rerank-english-v3.0"),
+    ...
+)
```
kdziedzic68, line 21: the same as with rephrasers

docs/how-to/document-search/use_reranker.md
```diff
@@ -0,0 +1,44 @@
+# How-To: Use Reranker
```
kdziedzic68, line 1: higher level comment: maybe we should consider moving rerankers to ragbits core ? i could see some potential use cases of them in other tasks too like eg. data augmentation

konrad-czarnota-ds: I think that this is a good topic for the grooming. This task was about creating docs to the components that are already there, they might be moved in the future but I wouldn't move functionalities between packages in doc creation task

kdziedzic68: agree - just wanted to raise a topic here. comment is not directly related to PR itself. marking as resolved

kdziedzic68 (approved):

PR #181, opened 2024-11-12
chore: docs for prompts

PatrykWyzgowski:
In this Pull Request I provide "how-to" for usage of `Prompt` class.

### DIFF
--- a/docs/how-to/use_prompting.md
+++ b/docs/how-to/use_prompting.md
@@ -98,11 +98,16 @@ The last two World Cup winners as of November 2017 are Germany (2014) and Spain 
 ```
 ### How to configure `Prompt`'s output data type
 #### Defining output as a Pydantic Model
-You can define the output of a prompt as a Pydantic model by specifying the output type as a generic parameter.
+You can define the output of a prompt as a Pydantic model by specifying the output type as a generic parameter. However, note that not all llm models support output schema definition. To use this feature effectively, you must set the `use_structured_output=True` flag when initializing the LLM. If this flag is not used, you will need to ensure that the JSON schema of your data model is incorporated into the prompt.
+
+Let’s revisit the previous example, making one adjustment: this time, we will define a structured output format.
+
 ```python
+import asyncio
 from pydantic import BaseModel
 
 from ragbits.core.prompt import Prompt
+from ragbits.core.llms.litellm import LiteLLM
 
 
 class QueryWithContext(BaseModel):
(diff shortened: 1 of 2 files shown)


docs/how-to/use_prompting.md
```diff
@@ -0,0 +1,210 @@
...
+
+    CONTEXT:
+    {% for item in context %}
+        {{ item }}
+    {% endfor %}
+    """
+
+...
+
+prompt = RAGPrompt(QueryWithContext(query=message, context=[i.get_text_representation() for i in results]))
```
kdziedzic68, line 78: what is message and results? cannot see the definition

docs/how-to/use_prompting.md
```diff
@@ -0,0 +1,210 @@
...
```
kdziedzic68, line 131: as above

docs/how-to/use_prompting.md
```diff
@@ -0,0 +1,210 @@
...
+class BooleanPrompt(Prompt[RoleInput, bool]):
+    user_prompt = "Are you {{ role }}? Answer 'yes' or 'no' only."
+
+
+boolean_prompt = BooleanPrompt(RoleInput(role = "an AI Assistant"))
+assert boolean_prompt.parse_response("true") is True
+```
+Please note that an actual response from LLM is going to raise `ResponseParsingError`. This is because the response is not natively parsable to boolean:
+```python
+llm = LiteLLM("gpt-4o-2024-08-06", use_structured_output=True)
```
kdziedzic68, line 157: i don't think that we should include non-working code here - only the correct example with description why do we need to implement parsing should be included.

docs/how-to/use_prompting.md
```diff
@@ -0,0 +1,210 @@
...
+        if "yes" in response.lower():
+            return True
+        elif "no" in response.lower():
+            return False
+        else:
+            raise ResponseParsingError("Response is not a valid boolean value.")
+```
+
+Another example of custom parser for handling empty responses.
+```python
```
kdziedzic68, line 194: i'm not sure why to include it ? pls show some use case

docs/how-to/use_prompting.md
```diff
@@ -0,0 +1,210 @@
...
+
+class CustomOutput:
+    def __init__(self, value: str):
+        self.value = value
+
+class CustomPrompt(Prompt[GreetingInput, CustomOutput]):
+    user_prompt = "Hello, {{ name }}!"
+
+    @staticmethod
+    def response_parser(response: str) -> CustomOutput:
```
kdziedzic68, line 206: typing should be changed if we allow response to be none

docs/how-to/use_prompting.md
```diff
@@ -0,0 +1,210 @@
...
+    user_prompt = """Tell me a joke."""
+```
+Passing the prompt to a model is as simple as:
+```python
+from ragbits.core.llms.litellm import LiteLLM
+
+
+llm = LiteLLM("gpt-4o-2024-08-06", use_structured_output=True)
+static_prompt = JokePrompt()
+print(await llm.generate(static_prompt))
```
kdziedzic68, line 33: for all the cases: console output should be also included below the code

kdziedzic68 (approved):

PR #184, opened 2024-11-12
feat: flatten the metadata structure in chroma vector store 

ds-sebastianchwilczynski:


micpst:
I think we should unflatten the metadata after retrieval from the vector store to preserve the original structure. Flattening is only for internal processing in chroma.

packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
```diff
@@ -165,17 +168,77 @@ async def list(
...
+            A flattened dictionary with primitive types and stringified complex types
+        """
+        items: dict[str, str | int | float | bool] = {}
+        for k, v in d.items():
+            new_key = f"{parent_key}{sep}{k}" if parent_key else k
+
+            if isinstance(v, dict):
+                items = {**items, **ChromaVectorStore._flatten_dict(v, new_key, sep=sep)}
+            else:
+                items[new_key] = v if isinstance(v, str | int | float | bool) else str(v)
```
micpst, line 204: We shouldn't store any complex types in the metadata, because we won't be able to reproduce them during unflattening, which will cause a validation error when parsing to the `Element` class. 

Support for `None` types should also be added. For example, if we convert the value to ``None`` and the field is typed as an optional string, we will incorrectly recreate the metadata.

pawel-chmielak-deepsense: Strings with `[]` in them will not be treated correctly later on - they'll be considered lists.

packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
```diff
@@ -165,17 +168,77 @@ async def list(
...
                 key=document,
                 vector=list(embedding),
-                metadata=metadata,  # type: ignore
+                metadata=self._unflatten_dict(metadata) if metadata else {},  # type: ignore
             )
             for id, metadata, embedding, document in zip(ids, metadatas, embeddings, documents, strict=True)
         ]
+
+    @staticmethod
+    def _flatten_dict(d: dict[str, Any], parent_key: str = "", sep: str = ".") -> dict[str, str | int | float | bool]:
```
pawel-chmielak-deepsense, line 185: This is such a generic functionality it could be considered an util function.

packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
```diff
@@ -165,17 +168,77 @@ async def list(
...
+            d: The dictionary to flatten
+            parent_key: The parent key for nested keys
+            sep: The separator between nested keys
+
+        Returns:
+            A flattened dictionary with primitive types and stringified complex types
+        """
+        items: dict[str, str | int | float | bool] = {}
+        for k, v in d.items():
+            new_key = f"{parent_key}{sep}{k}" if parent_key else k
```
pawel-chmielak-deepsense, line 199: If the separator is used in the key name, the flattening-unflattening will not work correctly.

packages/ragbits-core/tests/unit/vector_stores/test_chroma_utils.py
```diff
@@ -0,0 +1,59 @@
...
+def test_flatten_dict_with_non_dict_values():
+    """Test flattening a dictionary with various value types."""
+    input_dict = {
+        "key1": "value1",
+        "nested": {"subkey1": 42, "subkey2": [1, 2, 3], "subkey3": {"a": 1}, "subkey4": True, "subkey5": 3.14},
+    }
+    expected = {
+        "key1": "value1",
+        "nested.subkey1": 42,
+        "nested.subkey2": "[1, 2, 3]",
```
pawel-chmielak-deepsense, line 54: Have you considered the option below?
```
"nested.subkey2.0": 1,
"nested.subkey2.1": 2,
"nested.subkey2.2": 3,
```

You could also use a different separator for lists, if you prefer i.e.
```
"nested.subkey2:0": 1,
"nested.subkey2:1": 2,
"nested.subkey2:2": 3,
```

pawel-chmielak-deepsense (changes requested):

packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
mhordynski: moved to utils

packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
mhordynski: Ok, raising exception in that case - I think it's sane limitation to don't put such keys in vector stores

packages/ragbits-core/tests/unit/vector_stores/test_chroma_utils.py
mhordynski: I've added support for: 
"nested.subkey2[0]": 1,
"nested.subkey2[1]": 2,
"nested.subkey2[2]": 3,

packages/ragbits-core/src/ragbits/core/utils/dict_transformations.py
```diff
@@ -0,0 +1,145 @@
...
+                    items = {**items, **flatten_dict(item, list_key, sep=sep)}
+                else:
+                    items[list_key] = item if isinstance(item, str | int | float | bool) else str(item)
+        else:
+            items[new_key] = v if isinstance(v, str | int | float | bool) else str(v)
+
+    return items
+
+
+def unflatten_dict(input_dict: dict[Any, Any]) -> dict[Any, Any] | list:
```
pawel-chmielak-deepsense, line 46: `dict[Any, Any]`? Why not `dict[str, Any]` or `dict[str, str | int | float | bool | dict]`?

packages/ragbits-core/src/ragbits/core/utils/dict_transformations.py
```diff
@@ -0,0 +1,145 @@
...
+        - The function preserves the original values for non-nested keys
+        - Keys are sorted before processing to ensure consistent results
+
+    Attribution:
+        - This function is based on the answer by user "djtubig-malicex" on Stack Overflow: https://stackoverflow.com/a/67905359/27947364
+    """
+    new_dict: dict[Any, Any] = {}
+    field_keys = sorted(input_dict.keys())
+
+    def _parse_key(key: str) -> tuple[str | int | None, str | int | None]:
```
pawel-chmielak-deepsense, line 85: Returning `None, None` for a simple, flat key (e.g. `"key1"`) is an unusual design choice for a function called `_parse_key()`

packages/ragbits-core/src/ragbits/core/utils/dict_transformations.py
```diff
@@ -0,0 +1,145 @@
...
+    Raises:
+        ValueError: If the dictionary cannot be safely flattened due to the presence of the separator in the dict key.
+    """
+    items: dict[str, str | int | float | bool] = {}
+    for k, v in input_dict.items():
+        if sep in k:
+            raise ValueError(f"Separator '{sep}' found in key '{parent_key}' Cannot flatten dictionary safely.")
+
+        if k.endswith("]"):
+            raise ValueError(f"Key '{k}' cannot end with ']' Cannot flatten dictionary safely.")
```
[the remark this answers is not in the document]
mhordynski: even better, done

packages/ragbits-core/src/ragbits/core/utils/dict_transformations.py
mhordynski: fixed

packages/ragbits-core/src/ragbits/core/utils/dict_transformations.py
mhordynski: yeah, not a biggest fan of this, but it covers all the corner-cases and hopefully won't be touched ever again :sweat_smile: 

Renamed to _decompose_key which imo makes more sense in that scenario

PR #188, opened 2024-11-17
feat(core): streaming API for LLMs

kdziedzic68:


packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -80,12 +82,15 @@ async def generate_raw(
             Raw text response from LLM.
         """
         options = (self.default_options | options) if options else self.default_options
-
+        stream_output = False
+        if hasattr(prompt, "output_type"):
```
ludwiktrammer, line 86: # 1. 
The `output_type` is just a detail of how the `Prompt` class was implemented. `Prompt` is just one particular way to create prompts in dbally, we also support creating custom prompt classes (for example for prompts backed by an external source - like a database, YAML files, or an external API). That's way the LLM class here declares that it can take any `BasePrompt` object - we should stay true to that declaration. Streaming shouldn't be available exclusively to prompts created in this one particular way.

Moreover, you access `output_type` outside of the regular typing system. If we want prompt classes to be able to contain information whether the prompt should be streamed, we should explicitly declare a method like this in the abstract `BasePrompt` class, so that classes inheriting from that class can provide their implementation.

So, this is how I would implement it (if we really think that streaming should be declared on the level of the prompt - I have my doubts, as described in my other comment):

1. Add a new method to `BasePrompt`:
    ```python
    def is_streamed(self) -> bool:
        False
    ````
2. Add a more specific implementation of the new method to `Prompt`:
    ```python
    def is_streamed(self) -> bool:
        issubclass(self.output_type, tps.AsyncGeneratorType)
    ````
3. Replace like 93 here with `stream=prompt.is_streamed()` and remove lines 85-87

This way any prompt implementation (i.e., any class inheriting with `BasePrompt`) will be able to declare whether the prompt should be streamed, and code specific to how the `Prompt` class is implemented will stay inside `Prompt`.

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -80,12 +82,15 @@ async def generate_raw(
...
+            stream_output = prompt.output_type == tps.AsyncGeneratorType
```
ludwiktrammer, line 87: Does it really make sense to declare whether the response to the query should be streamed? On the face of it, it doesn't seem to me to me to be something directly connected to the content of the prompt. Instead, it seems to me that whether we want to stream or not is more of a run-time consideration - a function of things like whether the client that connected to our system supports streaming.

So for me, streaming doesn't seem like something that should be set statically in the prompt. Instead, I would see it as one of the LLMOptions.

packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -54,7 +54,7 @@ class BasePromptWithParser(Generic[OutputT], BasePrompt, metaclass=ABCMeta):
     """
 
     @abstractmethod
-    def parse_response(self, response: str) -> OutputT:
+    def parse_response(self, response: str | AsyncGenerator[str, None]) -> OutputT:
```
ludwiktrammer, line 57: Does it make sense to call `parse_response` for a streaming response?

packages/ragbits-core/src/ragbits/core/prompt/lab/app.py
```diff
@@ -112,7 +112,7 @@ def send_prompt_to_llm(state: PromptState) -> str:
     except Exception as e:  # pylint: disable=broad-except
         response = str(e)
 
-    return response
+    return response  # type: ignore
```
ludwiktrammer, line 115: Why? It seems to me that this will cause the app to crash for any streaming prompt - the problem that `# type: ignore` silences here is really a concern - it will cause `send_prompt_to_llm` to return a type different than what it declares and then it will cause gradio to try to set this non-string return value as a text content for a textbox.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -236,7 +236,7 @@ def parse_response(self, response: str) -> OutputT:
         Raises:
             ResponseParsingError: If the response cannot be parsed.
         """
-        return self.response_parser(response)
+        return self.response_parser(response)  # type: ignore
```
ludwiktrammer, line 239: I don't believe that sprinkling `type: ignore`s is usually the answer. We designed all those classes ourself. They should work properly with each other. If they don't after this changes, they should be redesigned to do (although it's still not clear to me what parsing a stream means).

ludwiktrammer (changes requested):

packages/ragbits-core/src/ragbits/core/llms/base.py
kdziedzic68: i was thinking about implementation of streaming as one of the `LLMOptions` however imo it is not LLM option and is related to the api rather than a model itself: LLM generates the answer token by token anyway - and we have access to the tokens - so the decision whether we want to wait for the generation to complete or read (anyway accesible pieces) on the fly is not related to the model directly

ludwiktrammer: It's related to how we want to use the model, isn't it? 

In simplest term, you model the decision as static and related to the particular prompt and to me it seems that the decision is often dynamic and related to the properties of the particular query that the system serves (e.g., the properties of the client that generated the query)

kdziedzic68: could agree with that - however we should than rename `LLMOptions` to `LLMClientOptions` and store there also info about output_schema and json_mode indicator.

btw in a case of` json_mode=True` we should also validate the input prompt

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
kdziedzic68: resolved by implementing separate method for streaming

packages/ragbits-core/src/ragbits/core/prompt/lab/app.py
kdziedzic68: resolved by implementing separate method for streaming

packages/ragbits-core/src/ragbits/core/prompt/base.py
kdziedzic68: resolved by implementing separate method for streaming

packages/ragbits-core/src/ragbits/core/llms/base.py
kdziedzic68: resolved by implementing separate method for streaming

packages/ragbits-core/src/ragbits/core/llms/base.py
kdziedzic68: resolved by implementing separate method for streaming

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -130,6 +130,32 @@ async def generate(
...
         return cast(OutputT, response)
 
+    async def generate_streaming(
+        self,
+        prompt: BasePrompt,
+        *,
+        options: LLMOptions | None = None,
+    ) -> AsyncGenerator[str, None]:
+        """
+        Prepares and sends a prompt to the LLM and returns the raw response (without parsing).
```
ludwiktrammer, line 140: Seems like the documentation copied from `generate_raw` instead of a description referring to this specific method

kdziedzic68: fixed

ludwiktrammer (approved):

PR #199, opened 2024-11-20
chore: install extras from ragbits index package

mackurzawa:
## Description

- Extras from subpackages are now included in the `ragbits` package.
- Modified `scripts/update_ragbits_package.py` to sync extras versions in `ragbits` with those declared in subpackages.

### DIFF
--- a/packages/ragbits/pyproject.toml
+++ b/packages/ragbits/pyproject.toml
@@ -39,6 +39,43 @@ dependencies = [
     "ragbits-core==0.3.0"
 ]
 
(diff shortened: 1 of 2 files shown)


mhordynski:
When selecting package to update as `ragbits-core` the extras logic is not triggered. 

Usually we run release on core, as it's updates all the packages which depend on it + updates `ragbits` package, so it's important to run extras logic also in this case.

mhordynski (approved):
lgtm

PR #205, opened 2024-11-26
feat(document-search): Option to choose between image and text embeddings for ImageElements

kdziedzic68:


packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -90,23 +90,35 @@ def from_vector_db_entry(cls, db_entry: VectorStoreEntry) -> "Element":
         """
         element_type = db_entry.metadata["element_type"]
         element_cls = Element._elements_registry[element_type]
+        if "embedding_type" in db_entry.metadata:
+            del db_entry.metadata["embedding_type"]
         return element_cls(**db_entry.metadata)
 
-    def to_vector_db_entry(self, vector: list[float]) -> VectorStoreEntry:
```
micpst, line 95: why is `embedding_type` optional here?

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -90,23 +90,35 @@ def from_vector_db_entry(cls, db_entry: VectorStoreEntry) -> "Element":
...
             The vector database entry
         """
+        metadata = self.model_dump(exclude={"id", "key"})
+        vector_store_entry_id = self.id
+        if embedding_type:
+            id_components = [
+                vector_store_entry_id,
+                str(embedding_type),
+            ]
+            vector_store_entry_id = str(uuid.uuid5(uuid.NAMESPACE_OID, ";".join(id_components)))
```
micpst, line 115: if we calculate the actual id here, we don't need the `id` attribute at class level

kdziedzic68: it is somehow different - here we compute id for vector store entry: eg. we have an image element that has description and ocr and image supporting embedding - those we want to create 2 entries in the store for embedding: one for descriptions, second for image and we want to distinguish between them - while Element itself remains the same

micpst: I get it, but this calculation is redundant since we use this id to distinguish vector store entries, not processor elements

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
micpst: I think embedding type should be enforced here

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
micpst: I'd even put this `id` computation into `VectorStoreEntry` class, but the implementation might be a bit shady since we'd need to rely on dynamic metadata field

kdziedzic68: but this computation does nothing with processor element - it does not modify it, so i've done it here as kind of workaroung of computing it while instantiation of `VectorStoreEntry` (we could ofc. have separate field embedding_type, however imo it is metadata field)

micpst: I'm just saying that calculating uuid5 twice for a vector entry id that is (currently) stored in an element is ambiguous, and everything should be calculated here or move it to the element class.

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
kdziedzic68: done with instantiation of `VectorStoreEntry` with class method. could have been `from_element` as well i guess, but than we would have some issues with nested imports.

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -150,16 +150,29 @@ async def insert_elements(self, elements: list[Element]) -> None:
...
+        elements_with_text = [element for element in elements if element.key]
+        images_with_text = [element for element in elements_with_text if isinstance(element, ImageElement)]
+        vectors = await self.embedder.embed_text([element.key for element in elements_with_text if element.key])
 
         image_elements = [element for element in elements if isinstance(element, ImageElement)]
-        entries = [element.to_vector_db_entry(vector) for element, vector in zip(elements, vectors, strict=False)]
+
+        num_images_with_no_textual_repr = len(image_elements) - len(images_with_text)
+        if num_images_with_no_textual_repr > 0:
+            warnings.warn(
```
ludwiktrammer, line 161: According to the ticket, this was supposed to be individual warnings for images that don't have ANY kind of embedding created.

As is, this will warn about ALL images in the scenario in which the user only want to create visual embeddings - which is a normal scenario, which we want to fully support.

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -42,15 +43,14 @@ def id(self) -> str:
         id_components = [
             self.document_meta.id,
             self.element_type,
-            self.key,
-            self.text_representation,
+            self.key or "null",
```
ludwiktrammer, line 46: Do `str(self.key)` to use Python's built-in `None` keyword instead of a magic string that isn't defined anyway and as such is quite brittle. If you really want to use the string "null", declare a constant - but personally I don't think that's necessary.

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -90,23 +90,27 @@ def from_vector_db_entry(cls, db_entry: VectorStoreEntry) -> "Element":
...
         return element_cls(**db_entry.metadata)
 
-    def to_vector_db_entry(self, vector: list[float]) -> VectorStoreEntry:
+    def to_vector_db_entry(self, vector: list[float], embedding_type: EmbeddingTypes) -> VectorStoreEntry:
         """
         Create a vector database entry from the element.
 
         Args:
             vector: The vector.
+            embedding_type: EmbeddingTypes.TEXT, EmbeddingTypes.IMAGE
```
ludwiktrammer, line 103: Listing specific `EmbeddingTypes` here is not a good practise.

1. What will happen when we add more values to `EmbeddingTypes`? Nobody will remember to update `to_vector_db_entry`'s documentation.
2. This is not a human-readable documentation of the parameter

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -90,23 +90,27 @@ def from_vector_db_entry(cls, db_entry: VectorStoreEntry) -> "Element":
...
 
         Returns:
             The vector database entry
         """
-        return VectorStoreEntry(
-            id=self.id,
-            key=self.key,
+        return VectorStoreEntry.from_element_data(
+            element_id=self.id,
+            element_key=self.key or "null",
```
ludwiktrammer, line 110: again - you can use `str(self.key)` instead of a magic string

ludwiktrammer (changes requested):

packages/ragbits-core/src/ragbits/core/embeddings/base.py
```diff
@@ -1,4 +1,14 @@
 from abc import ABC, abstractmethod
+from enum import Enum
+
+
+class EmbeddingTypes(Enum):
+    """
+    Enum for listing supported embedding types
```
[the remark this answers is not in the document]
kdziedzic68: done

packages/ragbits-core/src/ragbits/core/vector_stores/base.py
```diff
@@ -17,6 +19,29 @@ class VectorStoreEntry(BaseModel):
...
 
+    @classmethod
+    def from_element_data(
+        cls, element_id: str, element_key: str, vector: list[float], metadata: dict, embedding_type: EmbeddingTypes
+    ) -> "VectorStoreEntry":
+        """
+        Instantiates VectorStoreEntry from Element Data
+        Args:
+            element_id: str, id of element
+            element_key: str - text representation of element
```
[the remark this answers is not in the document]
kdziedzic68: this part of code has been removed

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -150,16 +150,29 @@ async def insert_elements(self, elements: list[Element]) -> None:
         Args:
             elements: The list of Elements to insert.
         """
-        vectors = await self.embedder.embed_text([element.key for element in elements])
+        elements_with_text = [element for element in elements if element.key]
+        images_with_text = [element for element in elements_with_text if isinstance(element, ImageElement)]
+        vectors = await self.embedder.embed_text([element.key for element in elements_with_text if element.key])
```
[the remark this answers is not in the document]
kdziedzic68: i was having issues with mypy - i can either do this or # type: ignore

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/_main.py
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
kdziedzic68: solved in a way discussed on a meeting

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -142,11 +148,37 @@ class ImageElement(Element):
...
+
+    @property
+    def id(self) -> str:
+        """
+        Retrieve the ID of the element, primarily used to represent the element's data.
+        The ID is computed as a UUID5 hash based on the element's data and embedding type
+        during insertion into the vector store.
+
+        Returns:
+           UUID: The computed UUID5 hash representing the element.
```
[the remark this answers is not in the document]
ludwiktrammer: nit: maybe we could have a separate function for returning the components that is used by `id` , so that inheriting classes can just add their values there, without having to repeat everything. 

So that for example `ImageElement` could simply contain this:

```
def get_id_components(self) -> str:
    components = super().components()
    return components + [hashlib.sha256(self.image_bytes).hexdigest()]
```

packages/ragbits-document-search/src/ragbits/document_search/_main.py
kdziedzic68: but isn't it a case that for example we define some processor that we intent to describe and / or ocr the image, but for some case it fails - api error, etc. shouldn't we somehow inform user about this case ?

packages/ragbits-core/src/ragbits/core/embeddings/base.py
```diff
@@ -1,4 +1,20 @@
 from abc import ABC, abstractmethod
+from enum import Enum
+
+
+class EmbeddingTypes(Enum):
```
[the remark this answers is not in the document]
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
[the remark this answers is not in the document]
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -28,29 +30,29 @@ class Element(BaseModel, ABC):
...
             self.document_meta.id,
             self.element_type,
-            self.key,
-            self.text_representation,
+            str(self.key),
+            str(self.text_representation),
             str(self.location),
         ]
-        return str(uuid.uuid5(uuid.NAMESPACE_OID, ";".join(id_components)))
+        return "-".join(id_components)
```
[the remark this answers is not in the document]
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
```diff
@@ -28,29 +30,29 @@ class Element(BaseModel, ABC):
...
-        Get the ID of the element. The id is primarly used as a key in the vector store.
-        The current representation is a UUID5 hash of various element metadata, including
-        its contents and location where it was sourced from.
+        Retrieve the ID of the element, primarily used to represent the element's data.
+        The ID is computed as a UUID5 hash based on the element's data and embedding type
+        during insertion into the vector store.
 
         Returns:
-            The ID in the form of a UUID5 hash.
+            UUID: The computed UUID5 hash representing the element.
```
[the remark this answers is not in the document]
kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/_main.py
ludwiktrammer: If there is an API error in processor than the processor should indicate that for the user. On the other hand, having a processor that doesn't add text descriptions for images is described in the ticket (and will be described in the documentation) as the main way of configuring the system to only create visual embedding for images.

kdziedzic68: ok will remove that however we also have a warning here for not supported vision embedding if the embedder does not support vision - would also remove it and warn only about not embedded at all images individually

kdziedzic68: done

packages/ragbits-document-search/src/ragbits/document_search/documents/element.py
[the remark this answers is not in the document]
kdziedzic68: done

ludwiktrammer (approved):

PR #206, opened 2024-11-27
fix(document-search): unbundle extra deps for sources

mhordynski:


packages/ragbits-document-search/src/ragbits/document_search/documents/sources.py
```diff
@@ -8,12 +9,16 @@
 from pydantic.alias_generators import to_snake
 from pydantic_core import CoreSchema, core_schema
 
+try:
+    from gcloud.aio.storage import Storage
+except ImportError:
+    contextlib.suppress(ImportError)
```
micpst, line 15: this should be either `pass` or refactored to `with` statement

micpst (approved):

PR #207, opened 2024-11-27
feat(document-search): add Ray processing strategy

pawel-chmielak-deepsense:


mhordynski:
Let's add how-to document about this feature, including how to submit ingestion as a job to a cluster

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
```diff
@@ -0,0 +1,47 @@
...
+            processor_router: The document processor router to use.
+            processor_overwrite: Forces the use of a specific processor, instead of the one provided by the router.
+
+        Returns:
+            A list of elements.
+
+        Raises:
+            ModuleNotFoundError: If Ray is not installed
+        """
+        import ray
```
kdziedzic68, line 39: why not to import at the begining of a file within try-except block? as it is done eg. with litellm or transformers modules within proper classes definitions? anyway we should also log some message like "in order to use blablabla... install the additional dependencies"

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
```diff
@@ -0,0 +1,47 @@
...
+
+        @ray.remote
+        def process_document_remotely(document: DocumentMeta | Document | Source) -> list[Element]:
+            return asyncio.run(self.process_document(document, processor_router, processor_overwrite))
+
+        tasks = [process_document_remotely.remote(document) for document in documents]
+
+        return sum(await asyncio.gather(*tasks), [])
```
kdziedzic68, line 47: why not to define some batch size?  as with local concurrent processing ?

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
```diff
@@ -0,0 +1,47 @@
...
+
+from ragbits.document_search.documents.document import Document, DocumentMeta
+from ragbits.document_search.documents.element import Element
+from ragbits.document_search.documents.sources import Source
+from ragbits.document_search.ingestion.document_processor import DocumentProcessorRouter
+from ragbits.document_search.ingestion.processor_strategies.base import ProcessingExecutionStrategy
+from ragbits.document_search.ingestion.providers.base import BaseProvider
+
+
+class DistributedProcessing(ProcessingExecutionStrategy):
```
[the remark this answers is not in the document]
pawel-chmielak-deepsense: I'd argue against it - that way we can later swap Ray for sth else without changing the API. Notice the extra requirements group is also called `distributed`, not `ray`.

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
pawel-chmielak-deepsense: Good idea, done!

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
pawel-chmielak-deepsense: Good idea, done!

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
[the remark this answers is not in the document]
kdziedzic68: ok - so i thought that maybe we could define some interface for this in that case, but can be left for later

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
kdziedzic68: so it is done the way that we have concurent processing in batch and concurrent processing of batches ?

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
```diff
@@ -0,0 +1,76 @@
...
+            processor_router: The document processor router to use.
+            processor_overwrite: Forces the use of a specific processor, instead of the one provided by the router.
+
+        Returns:
+            A list of elements.
+
+        Raises:
+            ModuleNotFoundError: If Ray is not installed
+        """
+        if not HAS_RAY:
```
kdziedzic68, line 56: maybe we should check it while instantiating the object ?

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
pawel-chmielak-deepsense: We split the load into batches and then send batches to different processes. So I'd say it's the latter.

packages/ragbits-document-search/src/ragbits/document_search/ingestion/processor_strategies/distributed.py
pawel-chmielak-deepsense: Done

kdziedzic68 (approved):

PR #210, opened 2024-11-28
feat(core): update document in vector store

mackurzawa:
## Description
This PR ensures that multiple ingestions of the same documents (in terms of `source.id`) do not retain `VectorStoreEntries` related to previous versions of the documents in the database. Before each ingestion, the `source.id` of the documents to be ingested is checked, and all `VectorStoreEntries` with the same `source.id` are removed from the database. The ingestion then proceeds as usual.

Implemented logic to check if the document's id is already present in `VectorStore`.

Implemented the `remove()` method for `ChromaVectorStore`, `QdrantVectorStore`, and `InMemoryVectorStore`.

Implemented 3 unit tests for each `VectorStore` to verify the functionality of the `remove()` methods.

Implemented 3 integration tests for each `VectorStore` with the following steps:
- Create two text documents, A and B, with content.
- Ingest documents A and B.
- Change the content of document B.
- Ingest document B.
- Verify that the database contains the content of document A.
- Verify that the database contains the new content of document B.
- Verify that the database does not contain the old content of document B.

## Testing
Testing can be done by running any of the integration tests:
`packages/ragbits-core/tests/integration/vector_stores/test_chroma.py`
`packages/ragbits-core/tests/integration/vector_stores/test_in_memory.py`
`packages/ragbits-core/tests/integration/vector_stores/test_qdrant.py`
or unit tests named in the below files:
`packages/ragbits-core/tests/unit/vector_stores/test_chroma.py`
`packages/ragbits-core/tests/unit/vector_stores/test_in_memory.py`
`packages/ragbits-core/tests/unit/vector_stores/test_qdrant.py`


packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
```diff
@@ -168,6 +184,10 @@ async def list(  # type: ignore
         Raises:
             MetadataNotFoundError: If the metadata is not found.
         """
+        collection_exists = await self._client.collection_exists(collection_name=self._index_name)
```
ludwiktrammer, line 187: When can this happen? Shouldn't this result in an exception?

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -141,8 +141,26 @@ async def ingest(
         elements = await self.processing_strategy.process_documents(
             documents, self.document_processor_router, document_processor
         )
+        await self.remove_entries_with_same_id(elements)
         await self.insert_elements(elements)
 
+    async def remove_entries_with_same_id(self, elements: list[Element]) -> None:
```
ludwiktrammer, line 147: maybe `remove_entries_with_same_sources`? Looking at the current name, I would expect this to be based on ids of the elements, not on the ids of sources

packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
mackurzawa: This can happen when we try to list elements from the Qdrant collection that has not yet had anything saved to it.

Every time we are ingesting new documents we need to check whether any elements from these documents already exist in the vector store. This procedure checks this every time - even when the vector store is empty and we want to do the ingestion for the first time.

Originally, without this check, Qdrant raised a ValueError saying:
`Collection {collection_name} not found`
Now, if the collection doesn't exist, it just returns an empty list.

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mackurzawa: Oh, great idea!

ludwiktrammer (approved):

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -141,8 +141,26 @@ async def ingest(
         elements = await self.processing_strategy.process_documents(
             documents, self.document_processor_router, document_processor
         )
+        await self.remove_entries_with_same_sources(elements)
         await self.insert_elements(elements)
 
+    async def remove_entries_with_same_sources(self, elements: list[Element]) -> None:
```
mhordynski, line 147: this method could be private

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -141,8 +141,26 @@ async def ingest(
...
+        """
+        Remove entries from the vector store whose source id is present in the elements' metadata.
+
+        Args:
+            elements: List of elements whose source ids will be checked and removed from the vector store if present.
+        """
+        unique_source_ids = {element.document_meta.source.id for element in elements}
+
+        ids_to_delete = []
+        for entry in await self.vector_store.list():
```
mhordynski, line 157: Shouldn't we use a where clause here to limit which elements we want to retrieve from the vector store? Fetching everything is not optimal.

Also, I noticed that in the Qdrant implementation, the list by default returns only 10 rows. We should probably change that to make this feature functional.

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -141,8 +141,26 @@ async def ingest(
...
+            if entry.metadata["document_meta"]["source"]["id"] in unique_source_ids:
```
mhordynski, line 158: It will fail when entry doesn't have any of these keys. We cannot assume that only records coming from document search are in vector store.

mhordynski (changes requested):

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mackurzawa: Regarding Qdrant limit: `AsyncQdrantClient.query_points()` takes `limit` argument which needs to be of `int` type. Default value of `limit` is set to 10 there, so it makes sense to pass `limit=limit or 10` when calling `query_points`, because limit is still set to 10 if not provided and it passes mypy checks. Maybe it's possible to ignore mypy there, but I haven't tested the functionality when `limit=None` is passed.

mackurzawa: Regarding listing all entries: we need to check what is the source of every entry in the vector store. I can't think of a better way to do this than by retrieving all the entries

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mackurzawa: Done

packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
```diff
@@ -146,6 +147,21 @@ async def retrieve(self, vector: list[float], options: VectorStoreOptions | None
...
 
+    @traceable
+    async def remove(self, ids: list[str]) -> None:
+        """
+        Remove entries from the vector store.
+
+        Args:
+            ids: The list of entries' IDs to remove.
+        """
+        await self._client.delete(
```
micpst, line 158: do you know what happens if the collection doesn't exist here? Will qdrant throw an exception or it's just noop?

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mackurzawa: Corrected!

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mackurzawa: Added checking if `limit` is `None` and if so, then setting `limit` to the number of the entries in the vector store.
Added `TODO` for passing `where` argument

packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
mackurzawa: Qdrant will throw a ValueError saying "`Collection {collection_name} not found`"

micpst: ok, probably we should introduce some unified error handling for vector databases. For now let's just mention it in the docstring that this method raises `ValueError`

packages/ragbits-core/tests/unit/vector_stores/test_chroma.py
```diff
@@ -101,6 +101,15 @@ async def test_retrieve(
         assert entry.key == result["content"]
 
 
+async def test_remove(mock_chromadb_store: ChromaVectorStore) -> None:
+    ids_to_remove = ["1c7d6b27-4ef1-537c-ad7c-676edb8bc8a8"]
+
+    await mock_chromadb_store.remove(ids_to_remove)
+
+    mock_chromadb_store._client.get_or_create_collection().delete.assert_called_once()  # type: ignore
+    mock_chromadb_store._client.get_or_create_collection().delete.assert_called_with(ids=ids_to_remove)  # type: ignore
```
pawel-chmielak-deepsense, line 110: IMO you should just list the ids available with `list` method instead of checking whether we call specific methods of a *private* field. This is just my opinion though, I'd love to hear about it from the rest of the team.

This remark applies to all unit tests.

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -141,8 +141,27 @@ async def ingest(
         elements = await self.processing_strategy.process_documents(
             documents, self.document_processor_router, document_processor
         )
+        await self.__remove_entries_with_same_sources(elements)
         await self.insert_elements(elements)
 
+    async def __remove_entries_with_same_sources(self, elements: list[Element]) -> None:
```
pawel-chmielak-deepsense, line 147: Is there a reason we use name mangling here, instead of marking it as private, as usual?

packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
mackurzawa: Docstring updated!

packages/ragbits-core/tests/integration/vector_stores/test_chroma.py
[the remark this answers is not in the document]
mackurzawa: I've chosen this approach, because unit tests are also divided based on type of vector store. But I can agree it would be less redundant and more readable, at least for now when these are the first integration tests. wdyt @mhordynski  ?

packages/ragbits-document-search/src/ragbits/document_search/_main.py
mackurzawa: No specific reason. I thought it'll be better, but can switch it to single `_`

mackurzawa: Ok, done

packages/ragbits-core/tests/integration/vector_stores/test_chroma.py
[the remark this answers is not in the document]
mhordynski: Agree with @pawel-chmielak-deepsense, when adding new VectorStore I would like to add just 1 line to parameterization call to test new implementation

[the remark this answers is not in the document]
mackurzawa: Refactored so everything is in one, parameterized test!

mhordynski (approved):
lgtm

PR #214, opened 2024-11-29
fix: context var using default mutable value

mhordynski:


packages/ragbits-core/src/ragbits/core/audit/base.py
```diff
@@ -13,12 +13,16 @@ class TraceHandler(Generic[SpanT], ABC):
     Base class for all trace handlers.
     """
 
+    _spans: ContextVar[list[SpanT]]
```
micpst, line 16: this is redundant attribute

micpst (approved):

PR #219, opened 2024-12-02
fix(document-search): do not create all of provider instances when importing document processor

kdziedzic68:


packages/ragbits-document-search/src/ragbits/document_search/ingestion/document_processor.py
```diff
@@ -106,7 +109,11 @@ def get_provider(self, document_meta: DocumentMeta) -> BaseProvider:
...
         """
-        provider = self._providers.get(document_meta.document_type)
-        if provider is None:
+        provider_class_or_provider = self._providers.get(document_meta.document_type)
+        if provider_class_or_provider is None:
             raise ValueError(f"No provider found for the document type {document_meta.document_type}")
+        elif isinstance(provider_class_or_provider, BaseProvider):
+            provider = provider_class_or_provider
+        else:
+            provider = provider_class_or_provider()
```
kdziedzic68, line 118: maybe after first instantiation we should cache an instance within provider dict ??

packages/ragbits-document-search/src/ragbits/document_search/ingestion/document_processor.py
```diff
@@ -7,30 +8,30 @@
 from ragbits.document_search.ingestion.providers.unstructured.images import UnstructuredImageProvider
 from ragbits.document_search.ingestion.providers.unstructured.pdf import UnstructuredPdfProvider
 
-ProvidersConfig = dict[DocumentType, BaseProvider]
+ProvidersConfig = dict[DocumentType, type[BaseProvider] | BaseProvider]
```
kdziedzic68, line 11: i'm not sure if we want to have distinct value type here, however - assuming later comment that we would potentially like to cache instantiated providers we would need both

ludwiktrammer: Suggestion(non-blocking): maybe instead of typing this as `type[BaseProvider] | BaseProvider` we could type it as `Callable[[], LLM] | BaseProvider`? This would allow us to accept all three: provider objects, provider classes, factory functions that return provider objects. And if I'm not mistaken, this could be made without doing any changes to the code logic, just the typing - the code itself seems to be already ready for factory functions.

ludwiktrammer: I think having an union of types here is okay.

On the other hand, I'm always nervous when using unions with dicts because dicts being [invariant](https://blog.daftcode.pl/covariance-contravariance-and-invariance-the-ultimate-python-guide-8fabc0c24278) causes them to behave in ways that are not always intuitive when it comes to unions (for example, forcing you to use `cast` in line 69, which normally wouldn't be necessary). So maybe we can think about changing `dict` to some specialised data class - but I don't think it's a priority right now

packages/ragbits-document-search/tests/integration/test_unstructured.py
```diff
@@ -33,7 +33,7 @@ async def test_document_processor_processes_text_document_with_unstructured_prov
 
     elements = await document_processor.get_provider(document_meta).process(document_meta)
 
-    assert isinstance(document_processor._providers[DocumentType.TXT], UnstructuredDefaultProvider)
+    assert isinstance(document_processor._providers[DocumentType.TXT], type(UnstructuredDefaultProvider))
```
ludwiktrammer, line 36: Adding some tests that would make sure that both kinds of configurations (type-based and instance-based) work correctly with `DocumentProcessorRouter` would be nice

kdziedzic68: done

ludwiktrammer (approved):

PR #232, opened 2024-12-04
feat(cli): add global -o / --output flag

kdziedzic68:


packages/ragbits-cli/src/ragbits/cli/app.py
```diff
@@ -0,0 +1,60 @@
...
+from pydantic import BaseModel
+from rich.console import Console
+from rich.table import Table
+
+
+@dataclass()
+class CliState:
+    """A dataclass describing CLI state"""
+
+    output_type: str = "text"
```
ludwiktrammer, line 15: It seems to me that `output_type` should be an enum

packages/ragbits-cli/src/ragbits/cli/app.py
```diff
@@ -0,0 +1,60 @@
...
+        self.state.output_type = output_type
+
+    def print_output(self, data: list[BaseModel]) -> None:
+        """
+        Process and display output based on the current state's output type.
+
+        Args:
+            data: list of ditionaries or list of pydantic models representing output of CLI function
+        """
+        first_el_instance = type(data[0])
```
ludwiktrammer, line 45: This should also handle the case when the list is empty

packages/ragbits-cli/src/ragbits/cli/app.py
```diff
@@ -0,0 +1,60 @@
...
+        if any(not isinstance(datapoint, first_el_instance) for datapoint in data):
+            raise ValueError("All the rows need to be of the same type")
+        data_dicts: list[dict] = [output.model_dump(mode="python") for output in data]
+        output_type = self.state.output_type
+        if output_type == "json":
+            print(json.dumps(data_dicts, indent=4))
+        elif output_type == "text":
+            table = Table(show_header=True, header_style="bold magenta")
+            for key in data_dicts[0]:
+                table.add_column(key.title())
```
ludwiktrammer, line 55: I would use `.model_json_schema()` from Pedantic instead. Each field there has a `title` field that by default is made to look human readable (for example "first_name" will be turned into "First Name", unlike when using `.title()`, which will result in "First_Name") and importantly can be controlled by the creator of the model by setting the `title` argument on the field (for example `first_name: str = Field(title='Name')`)

packages/ragbits-cli/src/ragbits/cli/app.py
```diff
@@ -0,0 +1,60 @@
...
+        Args:
+            output_type: str
+        Raises:
+            ValueError - if the output_type is not `json` or `text`
+        """
+        if output_type not in ["text", "json"]:
+            raise ValueError("Output type must be either 'text' or 'json'")
+        self.state.output_type = output_type
+
+    def print_output(self, data: list[BaseModel]) -> None:
```
ludwiktrammer, line 38: I feel it's likely that we will want this to become `print_output(self, data: list[BaseModel] | BaseModel)` and also to be able to take additional configuration regarding how things are displayed, but I think this can be left for the future.

ludwiktrammer: What I mean is that while `list[BaseModel]` should be displayed in a table maybe `BaseModel` could be displayed vertically by default instead - this would give us good support for both commands that list objects and commands that display details of a single object.

As to options - they could allow for more control of how things are laid out and which fields are included.

But as I said: this could become a separate ticket. We will know more when we start to add more commands.

packages/ragbits-cli/src/ragbits/cli/app.py
kdziedzic68: done

packages/ragbits-cli/src/ragbits/cli/app.py
kdziedzic68: done

packages/ragbits-cli/src/ragbits/cli/app.py
kdziedzic68: done

packages/ragbits-cli/src/ragbits/cli/app.py
kdziedzic68: ok - will be left for the future. for now i've just allowed `BaseModel` to be used as input

packages/ragbits-cli/src/ragbits/cli/app.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        Process and display output based on the current state's output type.
+
+        Args:
+            data: list of ditionaries or list of pydantic models representing output of CLI function
+        """
+        if isinstance(data, BaseModel):
+            data = [data]
+        if len(data) == 0:
+            raise ValueError("At last one element needs to be passed")
```
ludwiktrammer, line 59: In my mind an empty list is a normal use case that should be supported without raising an exception (that's what I meant by my previous comment - the previous version of the code would result in `KeyError`).

For example `ragbits vector-store list` and `ragbits vector-store query` may result in 0 results - that's entirely normal. I believe that in json mode this should print an empty list and in text mode - some non-error message (Like "No results")

packages/ragbits-cli/src/ragbits/cli/app.py
```diff
@@ -0,0 +1,76 @@
...
+            output_type: str
+        Raises:
+            ValueError - if the output_type is not `json` or `text`
+        """
+        if output_type not in ["text", "json"]:
+            raise ValueError("Output type must be either 'text' or 'json'")
+        if output_type == "json":
+            self.state.output_type = OutputType.JSON
+        if output_type == "text":
+            self.state.output_type = OutputType.TEXT
```
ludwiktrammer, line 47:
```suggestion
        Raises:
            ValueError - if the output_type is not one of the values in `OutputType`
        """
        self.state.output_type = OutputType(output_type)
```

packages/ragbits-cli/src/ragbits/cli/__init__.py
```diff
@@ -2,11 +2,25 @@
...
 
-app = Typer(no_args_is_help=True)
+from .app import CLI
+
+app = CLI(no_args_is_help=True)
+
+
+@app.callback()
+def output_type(
+    output: str = typer.Option("text", "--output", "-o", help="Set the output type (text or json)"),
```
ludwiktrammer, line 16: This should also be typed with the `OutputType` enum. See here: https://typer.tiangolo.com/tutorial/parameter-types/enum/

Otherwise the command will crash (generate a stack trace) for wrong values from user, instead of generating a user friendly message. Using an enum will also automatically document the possible values.

packages/ragbits-cli/src/ragbits/cli/app.py
ludwiktrammer: (although typing the callback with the `OutputType` enum will mean that `set_output_type` cal take an anum as well, making the problem go away altogether)

kdziedzic68: done - callback typed with the OutputType

packages/ragbits-cli/src/ragbits/cli/__init__.py
kdziedzic68: done - however it was not possible to make a typing exactly as in documentation.

packages/ragbits-cli/src/ragbits/cli/app.py
kdziedzic68: done

packages/ragbits-cli/src/ragbits/cli/__init__.py
```diff
@@ -2,11 +2,25 @@
...
+@app.callback()
+def output_type(
+    output: OutputType = typer.Option("text", "--output", "-o", help="Set the output type (text or json)"),  # noqa: B008
+) -> None:
+    """Sets an output type for the CLI
+    Args:
+        output: type of output to be set
+    """
+    app.set_output_type(output_type=output)
+    print(f"Output type set to {output}")
```
ludwiktrammer, line 23: I guess this is not intentional? Especially when outputting json it's important to be otherwise quiet, so that the json output can be parsed by other commands

kdziedzic68: this was for debuging purpose - removed

packages/ragbits-cli/src/ragbits/cli/__init__.py
```diff
@@ -1,12 +1,30 @@
...
+
+
+@app.callback()
+def output_type(
+    output: Annotated[
+        OutputType, typer.Option("--output", "-o", help="Set the output type (text or json)")
+    ] = OutputType.text.value,  # type: ignore
+) -> None:
+    # type ignore is a workaround for typer issue
+    # enforcing to use str as default value
```
ludwiktrammer, line 22:
```suggestion
    # `OutputType.text.value` used as a workaround for the issue with `typer.Option` not accepting Enum values
    output: Annotated[
        OutputType, typer.Option("--output", "-o", help="Set the output type (text or json)")
    ] = OutputType.text.value,  # type: ignore
) -> None:
```

The workaround is providing `OutputType.text.value` instead of the properly typed enum value. Type ignore is just shutting mypy up about it.

ludwiktrammer (approved):

PR #233, opened 2024-12-04
refactor: generalize creating objects from config

ludwiktrammer:


packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -32,7 +36,62 @@ def get_cls_from_config(cls_path: str, default_module: ModuleType) -> Any:  # no
...
+            An instance of the class initialized with the provided configuration.
+        """
+        subclass = get_cls_from_config(config.type, cls.default_module)
+        if not issubclass(subclass, cls):
+            raise InvalidConfigError(f"{subclass} is not a subclass of {cls}")
+
+        return subclass.from_config(config.config)
+
+    @classmethod
+    def from_config(cls, config: dict) -> Self:
```
kdziedzic68, line 87: maybe classes could have some (optional) class variable `config_model` or something and this should be a default behavior to validate the config here ?

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -73,24 +84,23 @@ def __init__(
         self.processing_strategy = processing_strategy or SequentialProcessing()
 
     @classmethod
-    def from_config(cls, config: dict) -> "DocumentSearch":
+    def from_config(cls, config: DocumentSearchConfig) -> "DocumentSearch":
```
kdziedzic68, line 87: isn't there an assumption that it should work on dictionaries or some dict configs ?

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
ludwiktrammer: Would you add a ticket about that to the backlog? I think it's an interesting idea to explore during grooming. But for now, I would leave that part as it was before, since the main purpose of this PR is to enable the work on CLI and the changes that are in this PR are enough for that.

packages/ragbits-document-search/src/ragbits/document_search/_main.py
ludwiktrammer: What do you mean by "assumption"?

ludwiktrammer: If you prefer, I can change it so it takes `dict` and then creates the `DocumentSearchConfig` object itself. It's just that I usually prefer specifically typed arguments to untyped arguments, but I can see the argument that `from_config` should be more consistent with `from_config` in other objects

ludwiktrammer: Changed - `DocumentSearchConfig.from_config()` now takes dict

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
kdziedzic68: added to backlog: https://github.com/deepsense-ai/ragbits/issues/237

kdziedzic68 (approved):

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -32,7 +36,62 @@ def get_cls_from_config(cls_path: str, default_module: ModuleType) -> Any:  # no
...
+
+        Args:
+            config: A model containing configuration details for the class.
+
+        Returns:
+            An instance of the class initialized with the provided configuration.
+        """
+        subclass = get_cls_from_config(config.type, cls.default_module)
+        if not issubclass(subclass, cls):
+            raise InvalidConfigError(f"{subclass} is not a subclass of {cls}")
```
micpst, line 82: please add this exception to the docstring

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -32,7 +36,62 @@ def get_cls_from_config(cls_path: str, default_module: ModuleType) -> Any:  # no
         except AttributeError as err:
             raise InvalidConfigError(f"Class {object_stringified} not found in module {module_stringified}") from err
 
+    if default_module is None:
+        raise InvalidConfigError("Given type string does not contain a module and no default module provided")
```
micpst, line 40: please update docstring

packages/ragbits-document-search/src/ragbits/document_search/_main.py
```diff
@@ -78,19 +89,20 @@ def from_config(cls, config: dict) -> "DocumentSearch":
...
-        query_rephraser = get_rephraser(config.get("rephraser"))
-        reranker = get_reranker(config.get("reranker"))
-        vector_store = get_vector_store(config["vector_store"])
-        processing_strategy = get_processing_strategy(config.get("processing_strategy"))
-
-        providers_config_dict: dict = config.get("providers", {})
-        providers_config = DocumentProcessorRouter.from_dict_to_providers_config(providers_config_dict)
+        model = DocumentSearchConfig.model_validate(config)
+
+        embedder = Embeddings.subclass_from_config(model.embedder)
```
micpst, line 99: please update docstring about potential exceptions raised from these builders

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
ludwiktrammer: done

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
ludwiktrammer: done

packages/ragbits-document-search/src/ragbits/document_search/_main.py
ludwiktrammer: done

PR #236, opened 2024-12-05
feat(core): install litellm by default

mhordynski:


ludwiktrammer (approved):

packages/ragbits-core/src/ragbits/core/embeddings/litellm.py
```diff
@@ -44,9 +39,6 @@ def __init__(
         Raises:
             ImportError: If the 'litellm' extra requirements are not installed.
         """
-        if not HAS_LITELLM:
-            raise ImportError("You need to install the 'litellm' extra requirements to use LiteLLM embeddings models")
```
micpst, line 48: update docstring pls, we no longer rise exception here

packages/ragbits-core/src/ragbits/core/llms/clients/litellm.py
```diff
@@ -68,9 +61,6 @@ def __init__(
         Raises:
             ImportError: If the 'litellm' extra requirements are not installed.
         """
-        if not HAS_LITELLM:
-            raise ImportError("You need to install the 'litellm' extra requirements to use LiteLLM models")
```
micpst, line 72: ditto

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -51,9 +46,6 @@ def __init__(
         Raises:
             ImportError: If the 'litellm' extra requirements are not installed.
         """
-        if not HAS_LITELLM:
-            raise ImportError("You need to install the 'litellm' extra requirements to use LiteLLM models")
```
micpst, line 55: ditto

mhordynski: good catch, thx

micpst (approved):

PR #238, opened 2024-12-05
fix typing for callback

kdziedzic68:


packages/ragbits-cli/src/ragbits/cli/__init__.py
```diff
@@ -14,15 +14,16 @@
 
 @app.callback()
 def output_type(
-    # `OutputType.text.value` used as a workaround for the issue with `typer.Option` not accepting Enum values
     output: Annotated[
-        OutputType, typer.Option("--output", "-o", help="Set the output type (text or json)")
-    ] = OutputType.text.value,  # type: ignore
+        OutputType | None, typer.Option("--output", "-o", help="Set the output type (text or json)")
```
ludwiktrammer, line 18: Why is this better? Having a default value defined here causes the auto generated documentation (`--help`)  to show it properly

PR #244, opened 2024-12-10
feat(vector-store): CLI commands for managing vector stores

ludwiktrammer:


packages/ragbits-cli/src/ragbits/cli/state.py
```diff
@@ -0,0 +1,64 @@
+import json
```
ludwiktrammer, line 1: Note: the code in this file is almost entirely copied from now deleted `cli/app.py`, with the main difference of the state/functions being standalone and using global state instead of being a part of a class inheriting from `typer.Typer`.

The only other difference is the typing of `data` argument for `print_output` changing from `list[BaseModel] | BaseModel` (mostly to avoid problems with `list` being invariant).

packages/ragbits-core/src/ragbits/core/config.py
```diff
@@ -19,5 +25,24 @@ class CoreConfig(BaseModel):
         LLMType.STRUCTURED_OUTPUT: "ragbits.core.llms.factory:simple_litellm_structured_output_factory",
     }
 
+    # Path to functions that returns instances of diffrent types of Ragbits objects
```
ludwiktrammer, line 28: Note: for LLMs `default_llm_factories` still exists as a separate setting - I don't love this but we have special needs for LLM factories, so I haven't come up with a better idea

packages/ragbits-core/src/ragbits/core/cli.py
```diff
@@ -1,104 +1,15 @@
...
-        if llm_factory is None:
-            raise ValueError("`llm_factory` must be provided")
-        llm: LLM = LLM.subclass_from_factory(llm_factory)
-
-        llm_output = asyncio.run(llm.generate(prompt))
-        response = LLMResponseCliOutput(question=prompt.chat, answer=llm_output)
-        app.print_output(response)
-
     app.add_typer(prompts_app, name="prompts", help="Commands for managing prompts")
+    app.add_typer(vector_stores_app, name="vector-store", help="Commands for managing vector stores")
```
kdziedzic68, line 15: i thought that maybe for the given module we should also auto-register sub module's cli's - however doesn't seem as sth very important

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+from ragbits.core.embeddings.base import Embeddings
+from ragbits.core.utils.config_handling import InvalidConfigError
+from ragbits.core.vector_stores.base import VectorStore, VectorStoreOptions
+
+vector_stores_app = typer.Typer(no_args_is_help=True)
+
+
+@dataclass
+class CLIState:
+    vector_store: VectorStore | None = None
```
kdziedzic68, line 21: in my opinion we should keep cli vector store config here not vector store itself

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+state: CLIState = CLIState()
+
+
+@vector_stores_app.callback()
+def common_args(
```
kdziedzic68, line 28: if we keep vector store config instead of vector store in the cli state - we could use this callback to eventually override the config - with no need to use it whenever we want to use any command

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
kdziedzic68: this way we would be able to keep some defaults here - or at last we should not use None as default, but instantiate some from defaults

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+        raise typer.Exit(1) from e
+
+
+@vector_stores_app.command(name="list")
+def list_entries(limit: int = 10, offset: int = 0) -> None:
+    """
+    List all objects in the chosen vector store.
+    """
+
+    async def run() -> None:
```
kdziedzic68, line 49: why do we use nested function here instead running vector store methods with `asyncio.run` ? i see no concurrency between task within this command

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+        if state.vector_store is None:
+            raise ValueError("Vector store not initialized")
+
+        entries = await state.vector_store.list(limit=limit, offset=offset)
+        print_output(entries)
+
+    asyncio.run(run())
+
+
+class RemovedItem(BaseModel):
```
kdziedzic68, line 59: shouldn't be a private class ? also - from pure aesthetical point of view i would move it to the top of the file

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+    id: str
+
+
+@vector_stores_app.command()
+def remove(ids: list[str]) -> None:
+    """
+    Remove objects from the chosen vector store.
+    """
+
+    async def run() -> None:
```
kdziedzic68, line 69: as one of above - is nested async function needed here ?

packages/ragbits-cli/src/ragbits/cli/state.py
```diff
@@ -0,0 +1,64 @@
...
+class CliState:
+    """A dataclass describing CLI state"""
+
+    output_type: OutputType = OutputType.text
+
+
+cli_state = CliState()
+
+
+def print_output(data: Sequence[BaseModel] | BaseModel) -> None:
```
kdziedzic68, line 28: in current state would rename it to `print_cli_output` or `cli_console_logger`

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+    k: int = 5,
+    max_distance: float | None = None,
+    embedder_factory_path: str | None = None,
+    embedder_yaml_path: str | None = None,
+) -> None:
+    """
+    Query the chosen vector store.
+    """
+
+    async def run() -> None:
```
kdziedzic68, line 94: here also no concurrent exeution within the  task , however thought that implementation of batched query on list of texts should be quiet straightforward

kdziedzic68: however maybe let's create backlog ticket for that

packages/ragbits-cli/src/ragbits/cli/__init__.py
```diff
@@ -23,12 +30,12 @@ def output_type(
     Args:
         output: type of output to be set
     """
-    app.set_output_type(output_type=output)
+    cli_state.output_type = output
 
 
-def main() -> None:
+def autodiscover() -> None:
```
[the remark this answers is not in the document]
ludwiktrammer: Ok, I don't have a preference here, will change to `autoregister`

packages/ragbits-cli/src/ragbits/cli/state.py
ludwiktrammer: Why? It's still under the `cli` namespace, but more importantly it simply prints to the standard output - just like Python's built-in `print` function. One could use it in non-cli contexts as wel.. On the other hand, "logger" has a very different connotation.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: I don't understand. Just like described in the ticket for this feature, you can use one of four ways to choose the store that you want to create:
1. Use the default store configuration from the project's `pyproject.yaml` (in this case ragbits will automatically create a store object from this configuration for you)
2. Use the default store factory function from the project's `pyproject.yaml` (in this case ragbits will automatically create a store object by calling the factory function)
3. Explicitly pass the filesystem path to a file containing store configuration (in this case ragbits will create a store object from this configuration)
4. Explicitly pass the Python path to a factory function (in this case ragbits will create a store object by calling the factory function).

This is the behavior that we agreed on during grooming. There is "no need to use it whenever we want to use any command" - the `factory_path` and `yaml_path` on the callback are optional and are used only when you want to override the defaults from the project's configuration for the given run of the command.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: This is simply my preference - this way you can use here the same syntax and conventions as with any other asynchronous code in the project. It's the same reason why in our example scripts you usually run `asyncio.run` on the entire `main` function instead of using it only with specific lines inside of `main`. Of course both would work.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: Some answer as above - I think running entire functions asynchronously and using the `async` / `await` syntax is more in-line with the rest of our code

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: This is already a private module - making things inside of it private wouldn't change anything.

When I declare utility stuff needed only for a single function I usually prefer to keep it near the function. But I can move it elsewhere if you want.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: I don't understand what could be made concurrent here

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+        if state.vector_store is None:
+            raise ValueError("Vector store not initialized")
+
+        try:
+            embedder = Embeddings.subclass_from_defaults(
```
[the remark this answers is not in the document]
ludwiktrammer: Changing our concept of vector store embed things themselves would be quite a big change to our current model  but obviously I'm okay with you creating such a ticket. On the other hand, I'm not sure it would meaningfully help with the potential problem scenario that you describe. Adding an embedder as a field of a `VectorStore` object wouldn't "store" the information - i.e., it would be as ephemeral as before, since the objects are created from configuration each time they are used. If you have mismatched configuration, you will have a problem in both cases. If you don't - you won't have a problem in neither.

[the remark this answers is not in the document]
kdziedzic68: we could accept list of texts instead of single text and define `async def run(text)`

and finally `asyncio.run(asyncio.gather(*[run(text) for text in texts]))`

i think that in this implementation it would be a very small change that would be potentially also benefitial (or tldr "low hanging fruit" :D)

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
kdziedzic68: ok - i forgot that callback is always run eventually with default arguments and had a misconception that if we want to run it we need to pass the arguments

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
kdziedzic68: ok had a misconception - it was related with the next comment about running callback

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
kdziedzic68: it is my preference - but if its yours it is not very important, i get yours

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
kdziedzic68: i mean that we could take a list of texts instead of single text as an input and write asynchronous nested run which takes text as an argument and process all the text concurrently - eventually which some batch size that could be another parameter. think that it is quiet low hanging fruit

packages/ragbits-cli/src/ragbits/cli/state.py
kdziedzic68: because `print_output` looks very similar to visually to `print` and when somebody - potentially fresh with the codebase could get confused while reading it

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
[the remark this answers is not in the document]
kdziedzic68: i think that to some point it would: if embedder would be part of a vector store than its name would specified within this config - if we add. eg. some embedder config hash - or even embeder name to the vector store name - than changing the embedder model would automatically also create a new vector store

packages/ragbits-cli/src/ragbits/cli/state.py
ludwiktrammer: It works just like `print` but prints structured output. Isn't that enough? I can rename it to `print_cli` if you would prefer it. But it's a commonly used utility function, I don't want to be a long java-ish name from 3 components.

packages/ragbits-core/src/ragbits/core/cli.py
ludwiktrammer: Personally, I wouldn't. We need autodiscovery on the package level because packages cannot rely on CLI being installed (since they don't depend on it). There is no similar problem on the inter-package level, so I think explicit (and fast) is better in this case than implicit and slower.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: More precisely callback can be run with empty arguments and the `subclass_from_defaults` method used inside of it is responsible for using the default in such case.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: Remember that this is a part of a CLI interface, so for humans to easily check stuff, and with its specific limitations. There is no data structures per se - just command line arguments as input, and writing text to standard output as output.

Right now one would use it like that:
```
ragbits vectore-store query "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas ac gravida justo, at venenatis neque. Nullam lacinia varius auctor. Praesent quis pretium augue."
```

And the output is a table of stuff from the data-store similar to the given text.

Of course I can change it so you can give an arbitrary number of texts separated by spaces, like this (although for me personally this doesn't look user-friendly):

```
ragbits vectore-store query "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Maecenas ac gravida justo, at venenatis neque. Nullam lacinia varius auctor. Praesent quis pretium augue." "raesent ante diam, elementum ut ligula nec, mattis vestibulum tellus. Etiam euismod, odio a varius interdum, arcu urna faucibus lorem, eu rhoncus sapien neque ut purus. " "Nulla facilisi. Interdum et malesuada fames ac ante ipsum primis in faucibus."
```

But what would be the output? Multiple tables - one table for each input text? That's not something that `print_output` currently supports, and I also think that would be confusing.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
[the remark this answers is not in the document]
ludwiktrammer: Currently, it's still hard for me to imagine - vector stores are typically external entities that we don't manage. We just connect to them based on the configuration that we get from the user. So there is no "automatically creating a new vector store" for us. It's like connecting to a remote SQL database. 

But if you have an idea how to do this, create a ticket and we will talk about it during grooming.

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
kdziedzic68: i would add task to discuss o grooming

kdziedzic68 (approved):

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+            factory_path_override=factory_path,
+            yaml_path_override=Path.cwd() / yaml_path if yaml_path else None,
+        )
+    except InvalidConfigError as e:
+        Console(stderr=True).print(e)
+        raise typer.Exit(1) from e
+
+
+@vector_stores_app.command(name="list")
+def list_entries(limit: int = 10, offset: int = 0) -> None:
```
mhordynski, line 44: list entries by default should not display vectors - it's not readable that way. Let's add a flag for that (--show-vectors? by default `false`)

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+state: CLIState = CLIState()
+
+
+@vector_stores_app.callback()
+def common_args(
+    factory_path: str | None = None,
+    yaml_path: str | None = None,
+) -> None:
+    try:
+        state.vector_store = VectorStore.subclass_from_defaults(
```
mhordynski, line 33: when there is now default factory and no factory_path / yaml_path arguments error message does not say anything meaningful other than `Could not find default factory or configuration for vector_store`

I think it should guide the user how to fix that

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
```diff
@@ -0,0 +1,117 @@
...
+        if cli_state.output_type == OutputType.text:
+            typer.echo(f"Removed entries with IDs: {', '.join(ids)}")
+        else:
+            print_output([RemovedItem(id=id) for id in ids])
+
+    asyncio.run(run())
+
+
+@vector_stores_app.command()
+def query(
```
mhordynski, line 83: Don't show vectors by default

mhordynski (changes requested):

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: We agreed to do this in separate PRs (but the same ticket)

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: We agreed to do this in separate PRs (but the same ticket)

packages/ragbits-core/src/ragbits/core/vector_stores/_cli.py
ludwiktrammer: We agreed to do this in separate PRs (but the same ticket)

mhordynski (approved):

PR #248, opened 2024-12-16
feat(core): Implement generic Options class

akonarski-ds:
This PR intoduces generic Options class that can be used with Embeddings, LLMs, VectoreStores that allows for convenient management of params passed to underlying APIs . It enables strong typing, but also allows for specifying extra params without changing core package code.


packages/ragbits-core/src/ragbits/core/vector_stores/base.py
```diff
@@ -31,16 +32,20 @@ class VectorStoreOptions(BaseModel, ABC):
     max_distance: float | None = None
 
 
-class VectorStore(WithConstructionConfig, ABC):
+VectorStoreOptionsType = TypeVar("VectorStoreOptionsType", bound=VectorStoreOptions)
+
+
+class VectorStore(Generic[VectorStoreOptionsType], WithConstructionConfig, ABC):
```
micpst, line 38: I wonder if we can abstract this generic class and create something like a `WithOptions` parent class that takes care of the templates and initialization of `default_options`.

micpst:
I don’t see any refactoring for the rerankers interface—was that intentional or just missed? It looks like they’ll need the new interface too.

packages/ragbits-core/src/ragbits/core/vector_stores/base.py
akonarski-ds: Hmm, I don't see a nice way to do it. Also, the initialization of `default_options` happens in `__init__` and is a one-liner. I don't think such simple code needs extra abstraction.

micpst: It'd make the interface more consistent. Right now, vector stores and embedders handle options differently - one uses class attributes, the other doesn’t - and it feels pretty confusing.

packages/ragbits-core/src/ragbits/core/embeddings/litellm.py
```diff
@@ -8,17 +8,31 @@
...
+class LiteLLMEmbeddings(Embeddings[LiteLLMEmbeddingsOptions]):
     """
     Client for creating text embeddings using LiteLLM API.
     """
 
     def __init__(
         self,
         model: str = "text-embedding-3-small",
-        options: dict | None = None,
+        options: LiteLLMEmbeddingsOptions | None = None,
```
ludwiktrammer, line 35: How would you use it when instantiating using with `Embeddings.from_config(...)` (note that the argument to `from_config` only contains simple types, serialized from an external yaml/json file)?

The same question to other classes.

See `LLM.from_config` to compare how `from_config` looks like for classes that used class based options before.

ludwiktrammer (changes requested):

packages/ragbits-core/src/ragbits/core/embeddings/litellm.py
akonarski-ds: Right, I'll add `from_config` that use class based option to `Embeddings` 

> The same question to other classes.

Which classes? 

Currently options will be supported in `LLMS`, `VectorStores` (already covered) and `Embeddings` (I need to add custom `from_config`) and `Rerankers` (in progress)

Do yo mean any other classes?

ludwiktrammer: What I meant is that it's a general issue that probably needs a general solution - `Embeddings` is an example, but I would try thinking of a solution that covers the current and feature cases. For example add an optional class property `options_cls` directly to `WithConstructionConfig` and change `WithConstructionConfig.from_config` to serialize `options` using `options_cls` if it is specified.

akonarski-ds:
@micpst @ludwiktrammer I've implemented the `ConfigurableComponent` class, which should address the concerns of both of you. Let me know what you think

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -169,3 +170,35 @@ def from_config(cls, config: dict) -> Self:
...
         """
         return cls(**config)
+
+
+class ConfigurableComponent(Generic[OptionsTypeVar], WithConstructionConfig):
+    """
+    Base class for components with configurable options.
+    """
+
+    _options_cls: type[OptionsTypeVar]
```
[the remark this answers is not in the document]
akonarski-ds: Second suggestion applied.

As for the first, generics in ClassVars are not supported: https://github.com/python/typing/discussions/1424

micpst (approved):

PR #265, opened 2024-12-30
Add setup_dev_env.fish

maciejklimek:


CONTRIBUTING.md
```diff
@@ -7,10 +7,16 @@ To build and run Ragbits from the source code:
 1. Requirements: [**uv**](https://docs.astral.sh/uv/getting-started/installation/) & [**python**](https://docs.astral.sh/uv/guides/install-python/) 3.10 or higher
 2. Install dependencies and run venv in editable mode:
 
+For bash users:
```
ludwiktrammer, line 10: That's not really true. I successfully use `venv/bin/activate` with `zsh` and I'm pretty sure that that the `activate` script works with almost all mainstream Linux shells. Looking at contents of  `venv/bin/activate*` it seems that the only exceptions are fish, and [nushell](https://www.nushell.sh/) (+ older Windows-specific / MacOS specific shells).

I wouldn't want to suggest that we only support bash and fish... and to be honest I'm a bit skeptical about making the README more complicated and having to support a new separate fish-specific setup script (which unlike the main one wouldn't be used and hence automatically tested in CI) for something so arbitrary (why fish?).

On the other hand... It's not clear to me why why need the two `source .venv/bin/activate` lines in `setup_dev_env.sh` at all. It's not needed for the `uv sync` command and it's certainly not needed at the end of the file. So I would change this PR to remove those two lines instead.

ludwiktrammer (changes requested):

CONTRIBUTING.md
mhordynski: >That's not really true. I successfully use venv/bin/activate with zsh and I'm pretty sure that that the activate script works with almost all mainstream Linux shells. Looking at contents of venv/bin/activate* it seems that the only exceptions are fish, and [nushell](https://www.nushell.sh/) (+ older Windows-specific / MacOS specific shells).

The correct terminology here would be:

`For POSIX-compliant shell users (e.g., bash, zsh):`

>On the other hand... It's not clear to me why why need the two source .venv/bin/activate lines in setup_dev_env.sh at all. It's not needed for the uv sync command and it's certainly not needed at the end of the file. So I would change this PR to remove those two lines instead.

The initial call to `.venv/bin/activate` can definitely be removed, as it is handled by `uv sync`. The idea behind the second call was to have the virtual environment activated automatically right after running this script. However, in my opinion, it is a bit unreliable. For instance, running `./setup_dev_env.sh` directly, rather than `source setup_dev_env.sh`, has no effect. I think we can drop this "feature".

mhordynski: Based on my previous comment and @micpst's take, I think it is the best to ditch the `setup_dev_env.sh` script. Since it just runs `uv sync`, we can do that straight away. As for the python version, it should be set with the `.python-version` file at the project root.

ludwiktrammer:
Closing since this has been fixed in #304

PR #267, opened 2025-01-08
feat: add support to pgvector

kzamlynska:


packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,351 @@
...
+        Args:
+            table_name: The name of the index.
+            db: The database connection string.
+            vector_size: The size of the vectors.
+            distance_method: The distance method to use.
+            default_options: The default options for querying the vector store.
+            metadata_store: The metadata store to use. If None, the metadata will be stored in pgVector db.
+        """
+        super().__init__(default_options=default_options, metadata_store=metadata_store)
+        conf = PgVectorConfig()
```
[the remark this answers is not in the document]
kzamlynska: removed

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,351 @@
...
+        if self.client:
+            async with self.client.acquire() as conn:
+                await conn.execute(create_vector_extension)
+                exists = await conn.fetchval(check_table_existence, self.table_name)
+
+                if not exists:
+                    create_command = self._create_table_command()
+                    await conn.execute(create_command)
+                    hnsw_name = self.table_name + "_hnsw_idx"
+                    query = create_index_query.format(
```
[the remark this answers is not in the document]
kzamlynska: I've added the check on table name in  __init__

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,351 @@
...
+from ragbits.core.metadata_stores.base import MetadataStore
+from ragbits.core.vector_stores.base import VectorStore, VectorStoreEntry, VectorStoreOptions, WhereQuery
+
+
+class PgVectorDistance:
+    """
+    Supported distance methods for pgVector.
+    """
+
+    DISTANCE_OPS = {
```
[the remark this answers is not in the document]
ludwiktrammer: I meant directly on the top-level of the module (in Python things doesn't have to be nested inside classes!). But this is also ok.

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,306 @@
...
+        # _table_name has been validated in the class constructor, and it is a valid table name.
+        query = f"SELECT * FROM {self._table_name}"  # noqa S608
+        if where:
+            filters = []
+            for key, value in where.items():
+                filters.append(f"{key} = {value}")
+            query += " WHERE " + " AND ".join(filters)
+
+        if limit is not None:
+            query += f" LIMIT {limit}"
```
ludwiktrammer, line 140: All those places where you add row values to SQL by f-string are places that create potencjal SQL-injection vulnerabilities. All values should be passed as parameters for SQL execution, all keys validated.

kzamlynska: I don't know how to do it here. It is query build on the conditions. it can be many of them. There can be LIMIT and OFFSET or not.

ludwiktrammer: When it comes to `LIMIT` and `OFFSET` specifically you have two options:

1. Always include limit and offset` in a parametrized form (something like `... LIMIT $1 OFFSET $2`) and where there is no limit set the limit value to `ALL`, and similarly where there is no offset set the offset value to `0`

2. Make sure that they are inits by passing the through the `int` constructor (for example `f" LIMIT {int(limit)}"`)

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,334 @@
...
+              str: sql query.
+        """
+        type_mapping = {
+            str: "TEXT",
+            list: f"VECTOR({self._vector_size})",
+            dict: "JSONB",
+        }
+        columns = []
+        type_hints = get_type_hints(VectorStoreEntry)
+        for column, column_type in type_hints.items():
```
ludwiktrammer, line 80: nit: this is not a bug but personally I'm not sure that why this is build dynamically. The fields on `VectorStoreEntry` are known, and you later reference those columns by name in your SQL queries, so I think they could be named explicitly here as well.  

Which means in my mind the entire method could simply become:

```
return f"CREATE TABLE {self._table_name} (id TEXT, key TEXT, metadata JSONB, vector VECTOR({int(self._vector_size)}));"
```

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,334 @@
...
+        query = f"SELECT * FROM {self._table_name}"  # noqa S608
+
+        values = []
+        index = 1
+
+        if query_options.max_distance and self._distance_method == "ip":
+            query += f""" WHERE vector ${index} '${index + 1}'
+            BETWEEN ${index + 2} AND ${index + 3}"""
+            values.extend([distance_operator, vector, (-1) * query_options.max_distance, query_options.max_distance])
+            index += 4
```
ludwiktrammer, line 116: nit: I would probably based those calculation on the length of `values` instead of keeping a separate `index` - less chance that the two desynchronize. But that's a personal preference.

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
kzamlynska: I remove the whole method. query is now static.

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,334 @@
...
+    def _create_table_command(self) -> str:
+        """
+        Create sql query for creating a pgVector table.
+
+        Returns:
+              str: sql query.
+        """
+        type_mapping = {
+            str: "TEXT",
+            list: f"VECTOR({self._vector_size})",
```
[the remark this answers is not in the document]
kzamlynska: corrected with removing the whole method

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -0,0 +1,310 @@
...
+                Not specifying the key means no filtering.
+            limit: The maximum number of entries to return.
+            offset: The number of entries to skip.
+
+        Returns:
+            sql query.
+        """
+        # _table_name has been validated in the class constructor, and it is a valid table name.
+        query = f"SELECT * FROM {self._table_name}"  # noqa S608
+
```
ludwiktrammer, line 124: Optional alternative implementation: (should work, but I did no testing):

```
query = f"SELECT * FROM {self._table_name} WHERE metadata @> $1 LIMIT $2 OFFSET $3"

values = [
    json.dumps(where) if where else "{}",
    limit or "ALL",
    offset or 0,
]

return query, values
```

ludwiktrammer: On further thought, `limit or "ALL"` probably won't work because `ALL` is a keyword, not a value.  

However, according to [the PostgreSQL documentation](https://www.postgresql.org/docs/current/queries-limit.html):

> `LIMIT ALL` is the same as omitting the `LIMIT` clause, as is `LIMIT` with a `NULL` argument.  

Since `NULL` is a value (represented in Python by `None`), this simpler version is likely to work:  

```python
query = f"SELECT * FROM {self._table_name} WHERE metadata @> $1 LIMIT $2 OFFSET $3"

values = [
    json.dumps(where) if where else "{}",
    limit,
    offset or 0,
]

return query, values
```

kzamlynska: I corrected it. It is easier to read that way.

mhordynski (approved):
lgtm, good job :medal_sports:

PR #271, opened 2025-01-09
Feat(conversations): Add last message recontextualizer

ludwiktrammer:
* I'll add the documentation in a separate PR (because landing the implementation itself in main is high priority)
* I'll add tests after #270 is merged (because they will involve creating a `MockLLM` class, which will be much easier after it is merged + doing it now will make merging it harder)

### DIFF
--- a/packages/ragbits-conversations/src/ragbits/conversations/history/compressors/llm.py
+++ b/packages/ragbits-conversations/src/ragbits/conversations/history/compressors/llm.py
@@ -1,0 +1,82 @@
+from pydantic import BaseModel
+
+from ragbits.conversations.history.compressors import ConversationHistoryCompressor
+from ragbits.core.llms.base import LLM
+from ragbits.core.prompt import ChatFormat, Prompt
+
+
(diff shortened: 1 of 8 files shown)


konrad-czarnota-ds (approved):
LGTM

packages/ragbits-conversations/src/ragbits/conversations/history/compressors/llm.py
```diff
@@ -0,0 +1,82 @@
...
+
+    async def compress(self, conversation: ChatFormat) -> str:
+        """
+        Contextualize the last message in the conversation history.
+
+        Args:
+            conversation: List of dicts with "role" and "content" keys, representing the chat history so far.
+                The most recent message should be from the user.
+        """
+        if len(conversation) == 0:
```
mhordynski, line 67: Maybe another check if user message is the only one (no history)? Then it just should return the exact message.

mhordynski:
Let's add `ragbits-conversations` to our index package `ragbits`

PR #274, opened 2025-01-15
feat: support metadata output from LLM

kdziedzic68:


packages/ragbits-core/src/ragbits/core/llms/base.py
micpst, line 1: I don't see the `generate_with_metadata` method that would return the logprobs

kdziedzic68: i changed a conception a bit: instead of generate_with_metadata `generate_raw` should return everything (with proper fields set in options

micpst: we lose this information in line 126 `response_str = cast(str, response["response"])` where we read only raw response from the LLM

kdziedzic68: yes i know - but in the case when we need logprobs - cannot we just call `generate_raw` and extract this information from there ?

kdziedzic68: ok - won't work if the output is pydantic model

kdziedzic68: done - is usable right now

micpst (approved):

PR #282, opened 2025-01-20
convert limits dict to httpx.Limits

kzamlynska:


packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
```diff
@@ -36,11 +38,13 @@ def __init__(
             distance_method: The distance metric to use when creating the collection.
             default_options: The default options for querying the vector store.
             metadata_store: The metadata store to use. If None, the metadata will be stored in Qdrant.
+            limits: The limits to apply to the HTTP client.
         """
         super().__init__(default_options=default_options, metadata_store=metadata_store)
         self._client = client
         self._index_name = index_name
         self._distance_method = distance_method
+        self.limits = limits
```
ludwiktrammer, line 47: It seems that this is not used (or needed)

kzamlynska: I've just removed it.

ludwiktrammer (approved):

PR #284, opened 2025-01-21
feat: add rerankers support

nczerep:


pyproject.toml
```diff
@@ -11,6 +11,7 @@ dependencies = [
     "ragbits-evaluate[relari]",
     "ragbits-guardrails[openai]",
     "ragbits-conversations",
+    "rerankers>=0.6.1",
```
mhordynski, line 14: It should be dependency of ragbits-document-search, not our workspace package

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/rerankers_answerdotai.py
```diff
@@ -0,0 +1,57 @@
+from collections.abc import Sequence
+
+from rerankers import Reranker as AnswerReranker
+
+from ragbits.core.audit import traceable
+from ragbits.document_search.documents.element import Element
+from ragbits.document_search.retrieval.rerankers.base import Reranker, RerankerOptions
+
+
+class AnswerDotAIRerankersReranker(Reranker[RerankerOptions]):
```
micpst, line 10: nit: a bit convoluted name, i'd suggest something like `AnswerAIReranker`.

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/rerankers_answerdotai.py
```diff
@@ -0,0 +1,57 @@
...
+    """
+    A [rerankers](https://github.com/AnswerDotAI/rerankers) re-ranker covering most popular re-ranking methods.
+    """
+
+    options_cls = RerankerOptions
+
+    def __init__(self, model: str, default_options: RerankerOptions | None = None, **options: str) -> None:
```
micpst, line 17: this `options` parameter is a bit confusing to me, since we use similar naming in the `rerank` method, i'd change it to `kwargs` or `rerankers_kwargs` to make it more explicit

packages/ragbits-document-search/src/ragbits/document_search/retrieval/rerankers/rerankers_answerdotai.py
```diff
@@ -0,0 +1,57 @@
...
+            query: The query to rerank the elements against.
+            options: The options for reranking.
+
+        Returns:
+            The reranked elements.
+        """
+        merged_options = (self.default_options | options) if options else self.default_options
+        documents = [element.text_representation for element in elements]
+
+        response = self.ranker.rank(
```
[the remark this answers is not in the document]
nczerep: you mean sth like `Raises:
        ValueError: If the query is empty or None.
        TypeError: If documents is not a list. ` ?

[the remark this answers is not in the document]
micpst: yes, exactly. i don't really know the `rerankers`, but if this lib sends any requests, there may be timeouts or rate limits on apis. we should catch these and map them to ragbits exceptions. for now, we can at least list them in the docstring so that someone working on exception handling later will have an easier time catching all the cases

micpst (approved):

PR #304, opened 2025-01-23
chore: remove the dev setup script

ludwiktrammer:


packages/ragbits-document-search/src/ragbits/document_search/ingestion/providers/base.py
```diff
@@ -13,7 +13,7 @@ class DocumentTypeNotSupportedError(Exception):
     """
 
     def __init__(self, provider_name: str, document_type: DocumentType) -> None:
-        message = f"Document type {document_type} is not supported by the {provider_name}"
+        message = f"Document type {document_type.value} is not supported by the {provider_name}"
```
ludwiktrammer, line 16: This tripped tests on some newer Python versions due to change in how enum values are displayed in newer Python version. That part makes sense. What I don't understand is why the problem only started to appear in CI in this PR.

.github/workflows/ci.yml
```diff
@@ -115,11 +108,10 @@ jobs:
 
       - name: Validate package build
         run: |
-          source .venv/bin/activate
           for dir in packages/*/; do uv build "$dir" --out-dir dist; done
 
       - name: Publish Package
-        uses: actions/upload-artifact@v3
+        uses: actions/upload-artifact@v4
```
ludwiktrammer, line 114: `upload-artifact@v3` will stop working in a week ([see the announcement](https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/)) but for some reason even today  some runs of the CI for this PR would fail with a message saying that `upload-artifact@v3` is not supported.

mhordynski (approved):
lgtm

PR #305, opened 2025-01-24
feat(document-search,cli): add command to ingest document

kdziedzic68:


packages/ragbits-document-search/tests/cli/test_ingest.py
```diff
@@ -0,0 +1,47 @@
...
+    [
+        ("*", 4),
+        ("test?.txt", 3),
+        ("test??.txt", 1),
+        ("test*.txt", 4),
+    ],
+)
+def test_ingest(pattern: str, num_expected: int) -> None:
+    runner = CliRunner(mix_stderr=False)
+    with tempfile.TemporaryDirectory() as temp_dir:
```
ludwiktrammer, line 31: nit: you should be able to use pytest's built-in [`tmp_path` fixture](https://docs.pytest.org/en/stable/how-to/tmp_path.html#the-tmp-path-fixture)

packages/ragbits-document-search/src/ragbits/document_search/cli.py
```diff
@@ -82,3 +82,19 @@ async def run() -> None:
...
+    source: Annotated[str, typer.Argument(help="Source pattern")],
+) -> None:
+    """
+    Ingest the elements from a given source to vector store.
+    """
+
+    async def run() -> None:
+        if state.document_search is None:
+            raise ValueError("Document search not initialized")
+        await state.document_search.ingest(source)
```
ludwiktrammer, line 98: nit: I don't have strong options on which behavior is better, but the `remove` command from vector-store prints a confirmation message, so now there will be some inconsistency in behavior between those two commands.

ludwiktrammer (approved):

PR #308, opened 2025-01-24
Kz/287 feat nested fields as columns

kzamlynska:
Adding possibility of having nested columns in output printed in a console.

Still working on unit tests for state


packages/ragbits-cli/src/ragbits/cli/state.py
```diff
@@ -64,23 +65,58 @@ def print_output_table(
...
+    Args:
+        column_name: name of the column to check
+        base_fields: model fields
+    """
+    fields = base_fields
+
+    *path_fragments, field_name = column_name.strip().split(".")
+    for fragment in path_fragments:
+        if fragment not in fields:
+            Console(stderr=True).print(f"Unknown column: {'.'.join(path_fragments + [field_name])}")
```
ludwiktrammer, line 105: nit: you could also optionally mention which part wasn't found, since you have the information anyway.

```
f"Unknown column: {'.'.join(path_fragments + [field_name])} ({fragment} not found)"
```

kzamlynska: I corrected this one.

packages/ragbits-cli/src/ragbits/cli/state.py
```diff
@@ -64,23 +65,58 @@ def print_output_table(
...
-            raise typer.Exit(1)
-
-        column = columns[key]
+    # check if columns are correct
+    for column_name in columns:
+        check_column_name_correctness(column_name, fields)
+        column = columns[column_name]
         if column.header == "":
-            column.header = titles.get(key, key)
+            column.header = titles.get(column_name, column_name).replace("_", " ").replace(".", " ").title()
```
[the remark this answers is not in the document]
ludwiktrammer: I mean we need titles, just from the correct model.

Maybe it's not clear what I want to achieve. Sorry about that. I'll try to explain:

Pydantic allows people to specify title of a given field. For example:

```
class Dog(BaseModel):
    name: str = Field(..., title="Puppy's name")
```

In the code that is currently in main we support this - we load those titles into the `titles` map, and use the column name **as a fallback**, only if the titles haven't been specified.

We don't want to lose this, but we need a better way of finding the titles (i.e., getting them from the right model). That's the reason I suggested changing `check_column_name_correctness` into a function that returns the actual field, checking the field for `title` and if it is there - using it. Autogenerating a title based on the column name should still remain a fallback.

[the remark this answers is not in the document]
kzamlynska: I think I understood and I corrected the code :)
So if there is a title of the filed I am taking just a title, if not I am taking the whole path leading to that field and reformat it a bit.
I even tested taking title if it exists.

packages/ragbits-cli/src/ragbits/cli/state.py
```diff
@@ -48,39 +49,74 @@ def print_output_table(
...
-    # Add headers to columns if not provided
-    for key in columns:
-        if key not in fields:
-            Console(stderr=True).print(f"Unknown column: {key}")
-            raise typer.Exit(1)
-
-        column = columns[key]
+    # check if columns are correct
+    for column_name in columns:
+        final_fields, field_name = check_column_name_correctness(column_name, base_fields)
```
ludwiktrammer, line 64: I would definitely change the name since now the purpose of the function is getting the field, not just checking. And you only need the single field, so you don't have to return multiple things. So I would change the function so that you can just write here:

```
field = get_nested_field(column_name, base_fields)
```

ludwiktrammer: Oh, and it would probably be good to make the function private (so `_get_nested_field`), so that it doesn't became part of our public interface

kzamlynska: You are right, I missed it. I corrected the method and change it to private one

ludwiktrammer (approved):

PR #309, opened 2025-01-27
Sparse vectors representation

nczerep:


packages/ragbits-core/src/ragbits/core/embeddings/sparse_vectors.py
```diff
@@ -0,0 +1,72 @@
...
+        if len(self.non_zero_dims) != len(self.non_zero_vals):
+            raise ValueError("It should be same number of non zero values and non zero positions")
+        if any(dim >= self.dim or dim < 0 for dim in self.non_zero_dims):
+            raise ValueError("Indexes should be in the range of the vector dim")
+
+    def __repr__(self) -> str:
+        return f"SparseVector(non_zero_dims={self.non_zero_dims}, non_zero_vals={self.non_zero_vals}, dim={self.dim})"
+
+
+class SparseEmbedding:
```
ludwiktrammer, line 24: This should be a `ConfigurableComponent` or at least have `WithConstructionConfig`, so that we can instantiate it from yaml / factory functions, just like all the other components.

packages/ragbits-core/src/ragbits/core/embeddings/sparse_vectors.py
```diff
@@ -0,0 +1,72 @@
...
+from dataclasses import dataclass
+
+
+@dataclass
+class SparseVector:
+    """Sparse Vector representation"""
+
+    non_zero_dims: list[int]
+    non_zero_vals: list[int]
+    dim: int
```
ludwiktrammer, line 12: nit: I think `length` would be more clear (and I'm not a fan of abbreviated property names) but I get that in large part this may depend on a personal viewpoint

ludwiktrammer (changes requested):

packages/ragbits-core/src/ragbits/core/embeddings/sparse_vectors.py
kdziedzic68: dimension vs length in terms of vectors i belive is kind of discourse between computer scientist and mathematician :)

packages/ragbits-core/src/ragbits/core/embeddings/sparse_vectors.py
kdziedzic68: done

mhordynski: > The ability to instantiate it from yaml / factory functions should be tested, similarly to how other configurable components are tested. 

^ tests are missing. Also why WithConstructionConfig and not ConfigurableComponent?

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
```diff
@@ -0,0 +1,70 @@
...
+        if len(self.non_zero_dims) != len(self.non_zero_vals):
+            raise ValueError("It should be same number of non zero values and non zero positions")
+        if any(dim >= self.dim or dim < 0 for dim in self.non_zero_dims):
+            raise ValueError("Indexes should be in the range of the vector dim")
+
+    def __repr__(self) -> str:
+        return f"SparseVector(non_zero_dims={self.non_zero_dims}, non_zero_vals={self.non_zero_vals}, dim={self.dim})"
+
+
+class SparseEmbedding(WithConstructionConfig, ABC):
```
mhordynski, line 28: It should be plural, to be consistent with our `Embeddings` class.

ludwiktrammer: Maybe it's time to rename the `Embeddings` class itself? It doesn't make much sense - it should be `Embedder` and `SparseEmbedder`. Even DocumenSearch when it uses the `Embeddings` class it calls it `embedder` (we have multiple instances of `embedder: Embeddings` in the code) - because that what it is. "Embeddings" would be an object containing actual embeddings, not a tool for creating them.

packages/ragbits-core/src/ragbits/core/embeddings/sparse_vectors.py
kdziedzic68: done. note that only instantiation from config has been tested to be consistent with dense embeddings tests

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
kdziedzic68: maybe not in this PR but in the one with VectorStore refactor ?

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
```diff
@@ -0,0 +1,98 @@
...
+    def __repr__(self) -> str:
+        return f"SparseVector(non_zero_dims={self.non_zero_dims}, non_zero_vals={self.non_zero_vals}, dim={self.dim})"
+
+
+class SparseEmbeddings(ConfigurableComponent[SparseEmbeddingsOptionsT], ABC):
+    """Sparse embedding interface"""
+
+    options_cls: type[SparseEmbeddingsOptionsT]
+    default_module: ClassVar = embeddings
+    configuration_key: ClassVar = "embedder"
```
ludwiktrammer, line 40: The key should be different since this is a different component with different interface, which cannot be used interchangeably with normal embeddings.

kdziedzic68: done

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
```diff
@@ -0,0 +1,98 @@
...
+class SparseVector:
+    """Sparse Vector representation"""
+
+    non_zero_dims: list[int]
+    non_zero_vals: list[int]
+    dim: int
+
+    def __post_init__(self) -> None:
+        if len(self.non_zero_dims) != len(self.non_zero_vals):
+            raise ValueError("It should be same number of non zero values and non zero positions")
```
ludwiktrammer, line 27:
```suggestion
            raise ValueError("There should be the same number of non-zero values as non-zero positions")
```

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
```diff
@@ -0,0 +1,98 @@
...
+    get_from_model_name: bool | None | NotGiven = NOT_GIVEN
+    min_token_count: int | None | NotGiven = NOT_GIVEN
+
+
+class BagOfTokens(SparseEmbeddings[BagOfTokensOptions]):
+    """BagofTokens implementations of sparse Embeddings interface"""
+
+    options_cls = BagOfTokensOptions
+
+    def __init__(self, encoding_or_model_name: str, default_options: BagOfTokensOptions | None = None):
```
ludwiktrammer, line 59: It's seems weird to me that the `get_from_model_name` option and the actual `encoding_or_model_name` are so separated. The `get_from_model_name` option controls how the `encoding_or_model_name` value is interpreted (so they have to be synchronized) but they are set in completely different places. 

I think you should be able to set them together. Have the default values for both in `BagOfTokensOptions` and be able to overwrite both when calling `embed_text`.

Or even better - just have a separate `encoding_name` and `model_name` options and validate that only one of them is set.

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
```diff
@@ -0,0 +1,98 @@
...
+        Transforms a list of texts into sparse vectors using bag-of-tokens representation.
+
+        Args:
+            texts: list of input texts.
+
+        Returns:
+            list of SparseVector instances.
+        """
+        vectors = []
+        min_token_count = self.default_options.min_token_count or -np.inf
```
ludwiktrammer, line 84:
```suggestion
        min_token_count = self.default_options.min_token_count or float("-inf")
```

I don't think there is a reason to bring whole numpy here just for this one value

packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
kdziedzic68: done

mhordynski (approved):

PR #319, opened 2025-01-28
docs:  how to on tracing

kzamlynska:


docs/how-to/use_tracing.md
```diff
@@ -0,0 +1,106 @@
+# How to use tracing in ragbits
+
+Ragbits provides two options for trace handling: [**OpenTelemetry**](https://www.ibm.com/products/instana/opentelemetry)
```
mhordynski, line 3: https://opentelemetry.io/

docs/how-to/use_tracing.md
```diff
@@ -0,0 +1,106 @@
...
+and **CLI**.
+To use either of these trace handlers, you need to import the traceable decorator from ragbits.core.audit
```
mhordynski, line 5: I think this should go in different section, something like "How to trace your own code with ragbits". And there we should describe both `@traceable` decorator and direct traces through trace context manager.

docs/how-to/use_tracing.md
```diff
@@ -0,0 +1,106 @@
+# How to use tracing in ragbits
```
mhordynski, line 1: I think we should start with explaining that each ragbits component is traced and users can out-of-the-box start gathering useful insights about their application. 

I would rather avoid statements like "two options", "both methods", etc. as audit is prepared to provide more than two handlers and new ones are very likely to be developed in the future

docs/how-to/use_tracing.md
```diff
@@ -0,0 +1,106 @@
...
+    -p 16686:16686 \
+    -p 4317:4317 \
+    jaegertracing/all-in-one:1.62.0
+```
+2. **Open the Jaeger UI in your browser**:
+
+```
+http://localhost:16686
+```
+To check the OpenTelemetry trace handler you can also run our example:  ```examples/document-search/otel.py```
```
mhordynski, line 106: Let's have link to github here

docs/how-to/use_tracing.md
kzamlynska: changed

docs/how-to/use_tracing.md
kzamlynska: I moved this to different section. I am not sure how to use  direct traces through trace context manager.

docs/how-to/use_tracing.md
kzamlynska: I've changed that. I linked this to github path, but to make it more visible I left text short.

docs/how-to/use_tracing.md
kzamlynska: added the description `with trace`

docs/how-to/core/use_tracing.md
```diff
@@ -0,0 +1,133 @@
...
+   import trace from ragbits.core.audit and run a part of code you wish to trace within `with trace()` context:
+```python
+from ragbits.core.audit import trace
+
+def add_more_numbers(a: int, b: int, c: int) -> int:
+    with trace(name='check a', first_number=a) as outputs:
+        if a % 2 == 0:
+           result = a + 1 +c
+           outputs.first_result = result
+        with trace(name='check b', second_number=b) as outputs:
```
mhordynski, line 36: new outputs overrides previous one, which is misleading. Let's name them outputs_a, outputs_b

Also let's have `outputs_a.result= result` rather than first/second_result

docs/how-to/core/use_tracing.md
```diff
@@ -0,0 +1,133 @@
...
+                result = a + 2 + c
+                outputs.second_result = result
+```
+   In the tracing tree, inputs are provided as keyword arguments when initializing the trace context,
+   while outputs must be explicitly assigned within the traced block.
+
+
+You can enable the desired trace handler in your script using the following method:
+```audit.set_trace_handlers(trace_handler_name)```
+Keep in mind that you can use more than one trace handling both methods in one script.
```
mhordynski, line 47:
```suggestion
Keep in mind that you can use more than one trace handling method simultaneously.
```

docs/how-to/core/use_tracing.md
kzamlynska: corrected

mhordynski (approved):

PR #321, opened 2025-01-29
chore: stricter mkdocs validation

ludwiktrammer:
Before some problems in mkdocs that should be reported as warnings were reported as "info" instead and as such not reported during our CI process.

This commit uses the configuration that mkdocs documentation describes as "recommended settings for most sites": https://www.mkdocs.org/user-guide/configuration/#validation


kzamlynska (approved):
It is working and it is helpfull.

PR #324, opened 2025-01-30
demo: conversational pipeline - implementation 1

ludwiktrammer:


mhordynski:
@ludwiktrammer that's rad, I really like how it looks :tada:
I'll post couple of my thoughts in threads to keep them apart.

mhordynski:
I think ConversationPipeline should have base class with common interface.

Thiw will allow users to have their fully custom pipeline, just ensuring that input / output can be handled by other ragbits components. This will allow us to easily expose any predefined / user custom-defined conversation pipeline through common API (for example to connect it with our demo frontend app).

mhordynski:
I'm wondering if we can simplify interface of the plugins. 

Right now they can serve two main functionalities:  
* process state
* process output

What if we could say that plugin just have one method called `run` (or something like that) and defines it's input and output. In that case for ConversationPipeline we would have plugins with interfaces:
```
ConversationPipelineState -> ConversationPipelineState
ConversationPipelineResult -> ConversationPipelineResult
```

and probably if we say that StateToChat / LLM are also on the same level then we'll have

```
ConversationPipelineState -> Prompt
Prompt -> ConversationPipelineResult
```

micpst:
looks interesting, wonder if we can reduce the boilerplate code to the bare minimum here

micpst:
perhaps we should ditch this conversation package, and focus on the bigger picture and have some sort of generic computation interface for agent workflows (not necessarily for conversations, but something more generic)

i personally really like [pokcetflow](https://github.com/the-pocket/PocketFlow) it’s just 100 lines of code and it’s super clean and intuitive, maybe if we can have something similar in ragbits

PR #329, opened 2025-02-03
feat: add azure storage source

kzamlynska:


mhordynski:
please rename PR to comply with conventional commits

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
```diff
@@ -0,0 +1,176 @@
...
+    blob_name: str
+    _blob_service: BlobServiceClient | None = None
+    account_name: str | None = None
+
+    @property
+    def id(self) -> str:
+        """
+        Get the source ID, which is the full blob URL.
+        """
+        return f"{self.container_name}/{self.blob_name}"
```
mhordynski, line 31: unique identifier is missing to figure out by id if source is from azure. 
Also account_name should be included here.

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
```diff
@@ -0,0 +1,176 @@
...
+class AzureBlobStorage(Source):
+    """
+    An object representing an Azure Blob Storage dataset source.
+    """
+
+    protocol: ClassVar[str] = "azure"
+    container_name: str
+    blob_name: str
+    _blob_service: BlobServiceClient | None = None
+    account_name: str | None = None
```
mhordynski, line 24: Why account_name can be none? I think it should be always there.

pyproject.toml
```diff
@@ -12,6 +12,8 @@ dependencies = [
     "ragbits-evaluate[relari]",
     "ragbits-guardrails[openai]",
     "ragbits-conversations",
+    "azure-storage-blob>=12.24.1",
```
mhordynski, line 15: this dependencies should not be here:

1. You should modify document-search package rather than workspace one.
2. These shouldn't be required always, let's keep them in extras.

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
kzamlynska: I've added it. The account name can be provided as an class key arguments or through environemtnal variable

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
kzamlynska: I've changed the code: it still can be none if it is provided as environmental variable.

pyproject.toml
kzamlynska: I moved dependecies to optional in uv.lock and added it to the pyproject in document-search.

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
```diff
@@ -0,0 +1,204 @@
...
+    container_name: str
+    blob_name: str
+    _blob_service: BlobServiceClient | None = None
+
+    @model_validator(mode="after")
+    def set_account_name(self) -> Self:
+        """
+        Validates that the account name is set.
+        """
+        self.account_name = self.account_name or os.getenv("AZURE_STORAGE_ACCOUNT_NAME")
```
mhordynski, line 34: I don't get still how this env should work. Potentially we can have different sources from different accounts in one DocumentSearch instance.

Let's say:

we have files `dogs.md`, `cats.md`,`home.md` from account `animals` container `english`
but we also have `python.md`, `java.md`, `home.md` from account `programming` container `test`

Having this environment variable would cause weird behaviour that depending on the running system configuration we would either index `home.md` from animals or programming account. IMO let's be here explicit and always require account_name to be a part of Source initialization params.

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
```diff
@@ -0,0 +1,204 @@
...
+from azure.storage.blob import BlobServiceClient
+from pydantic import model_validator
+from typing_extensions import Self
+
+from ragbits.core.utils.decorators import requires_dependencies
+from ragbits.document_search.documents.exceptions import SourceConnectionError, SourceNotFoundError
+from ragbits.document_search.documents.sources import Source, get_local_storage_dir
+
+
+class AzureBlobStorage(Source):
```
micpst, line 18: nit: to keep the naming consistent with other implementations, i'd change the name to `AzureBlobStorageSource`

packages/ragbits-document-search/pyproject.toml
```diff
@@ -50,7 +50,11 @@ huggingface = [
 distributed = [
     "ray>=2.39.0",
 ]
-
+azure=[
+    "azure-storage-blob>=12.24.1",
```
micpst, line 54: please make these requirements version compatible, in case azure breaks api in the future

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
kzamlynska: I've just checked that to have an access to different accounts I need to have user name. Should I change the code so it will work for many accounts?

packages/ragbits-document-search/src/ragbits/document_search/documents/azure_storage_source.py
kzamlynska: I've  changed that

packages/ragbits-document-search/pyproject.toml
kzamlynska: done

mhordynski (approved):
looks good, please merge #351 first and then adapt this PR to new structure

PR #330, opened 2025-02-04
demo: conversational pipeline (+generic pipleline) - implementation 2

ludwiktrammer:


packages/ragbits-core/src/ragbits/core/pipelines/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    @abstractmethod
+    async def run(self, /, input_data: InputT) -> OutputT:
+        """
+        Run the step with the given input data
+        """
+
+
+class Pipeline(Step[InputT, OutputT]):
```
ludwiktrammer, line 24: Note: Every `Pipeline` is also a `Step. Which means:

1. You can nest pipelines (add pipelines as steps of other pipelines)
2. Most likely, typing an argument as `Pipeline[input type, output type]` will never make sense - it's always better to accept any `Step[input type, output type]`. This way both `Pipeline` and `Step` can be given as the value for the argument, so users can freely choose whether they want to involve an entire pipeline or do everything in one step.

(2) also probably suggest that "Step" is not the based name. Because really `Step` can be just anything that can take some (typed) input, run some logic on it, and return some (typed) output. And `Pipeline` is just a way to chain multiple of such objects.

packages/ragbits-core/src/ragbits/core/pipelines/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    A pipeline that consists of multiple steps that process input data and return output data
+    """
+
+    def __init__(self, *steps: Step) -> None:
+        self.steps = steps
+        self._validate_intrastep_types()
```
ludwiktrammer, line 31: Note: during initialization we validate that individual steps have types that are compatible with each other (i.e, that the return type of each step is within bounds of the input type of its following step), **but** due to specifics of the Python generics system, we can validate that the steps have types compatible with the pipeline itself only when the pipeline is run.

packages/ragbits-core/src/ragbits/core/pipelines/pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        Check if the actual type is a subtype of the needed type.
+        """
+        # Covert new union type syntax to Union objects, as typing_utils.issubtype
+        # does not support the new syntax
+        if isinstance(needed_type, UnionType):
+            needed_type = Union[tuple(needed_type.__args__)]  # noqa: UP007
+        if isinstance(actual_type, UnionType):
+            actual_type = Union[tuple(actual_type.__args__)]  # noqa: UP007
+        return issubtype(needed_type, actual_type)
```
ludwiktrammer, line 44: `issubtype` comes from an external library (`typing_utils`) and unfortunately it is on the library's author to implement all the logic behind different typing hints in Python.

Which means that it usually works, but in some cases (new additions to the typing system, less know types, edge cases, generics) may not. For example, the library don't support new pipe syntax for unions - that's why I had to convert from new to old syntax above.

I see this as a problem, because if we go with this implementation, it will _look_ like Steps and Pipelines can be typed with any typing hints available, and usually it indeed work. But for some more advanced hints it will just not.

packages/ragbits-conversations/src/ragbits/conversations/piepline/pipeline.py
```diff
@@ -0,0 +1,70 @@
...
+    async def run(self, state: ConversationPipelineState) -> ConversationPipelineResult:
+        """
+        Processes the conversation pipeline state and returns the updated result.
+        """
+        chat = await self.state_to_chat.run(state)
+        stream = self.llm.generate_streaming(chat)
+        return ConversationPipelineResult(plugin_metadata=state.plugin_metadata, output_stream=stream)
+
+
+class ConversationPiepline(Step):
```
ludwiktrammer, line 29: Note: `ConversationPiepline` is meant to be easier to use (than a generic `Pipeline`) with conversation as it requires specific arguments and types, showing the user exactly what they need to provide to create a conversational pipeline.

But it is compatible with `Pipeline` in many different ways:
1. You can pass `ConversationPiepline` object as a step to any `Pipeline`
2. You can pass any `Pipeline` as a `ConversationPiepline` step (preprocessor, postprocessor, state_to_chat)
3. Places that need a conversation pipeline will accept anything typed as `Step[ConversationPipelineState | str, ConversationPipelineResult]`, so it means that any such place will accept any of those:
    - A `ConversationPiepline`
    - A `Pipeline`
    - A custom `Step` implementation
4. Step implementations are compatible between `ConversationPiepline` and `Pipeline`
5.  `ConversationPiepline` uses a `Pipeline` behind the scenes

(Every time the word "any" is used within the list above it's a shortcut, it should really be "any as long as the types match")

packages/ragbits-conversations/src/ragbits/conversations/piepline/pipeline.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    Class that runs a conversation pipeline with the given plugins
+    """
+
+    def __init__(
+        self,
+        llm: LLM,
+        preprocessors: Sequence[Step[ConversationPipelineState, ConversationPipelineState]] = [],
```
ludwiktrammer, line 37: Note: in this implementation you don't have plugins that can simulanously act on different stages of the pipeline (like you do with implementation 1). Here, if supporting a functionality requires both pre- and post- processing, you will need to have two different implementations and pass both as objects.

PR #332, opened 2025-02-04
feat: multi-query retrieval

akotyla:


docs/how-to/document_search/use_rephraser.md
```diff
@@ -57,6 +57,48 @@ class QueryRephraserPrompt(Prompt[QueryRephraserInput, str]):
...
+    rephrased = await rephraser.rephrase("What time is it?")
+    print(rephrased)
+
+asyncio.run(main())
+```
+
+The console should print multiple variations:
+
+```python
+['What time is it?', 'Can you tell me the time?', 'What's the current time?']
```
mhordynski, line 99: last item probably would break formatting due to ` character. Maybe let's switch to double quotes here

akotyla: right, fixed

PR #339, opened 2025-02-10
feat:  aws source implementation

kzamlynska:
AWS S3 Source implemented


packages/ragbits-document-search/src/ragbits/document_search/documents/aws_s3_source.py
```diff
@@ -0,0 +1,183 @@
...
+
+import boto3
+from botocore.client import BaseClient
+from botocore.exceptions import ClientError, NoCredentialsError, PartialCredentialsError
+
+from ragbits.core.utils.decorators import requires_dependencies
+from ragbits.document_search.documents.sources import Source, get_local_storage_dir
+
+
+class AWSSource(Source):
```
micpst, line 14: nit: just to keep naming consistent, i'd change the name to `S3Source`. we have `GCSSource`, which uses the name of the GCP storage service name, here i'd also expect the name of the AWS storage service, which is S3 instead of the name of the cloud provider

packages/ragbits-document-search/src/ragbits/document_search/documents/aws_s3_source.py
```diff
@@ -0,0 +1,183 @@
...
+             Exception: If another error occurs.
+        """
+        if cls._s3_client is None:
+            try:
+                cls._s3_client = boto3.client("s3")
+                cls._s3_client.head_bucket(Bucket=bucket_name)  # This triggers a credentials check
+            except (NoCredentialsError, PartialCredentialsError) as e:
+                raise RuntimeError("AWS credentials are missing or incomplete. Please configure them.") from e
+            except ClientError as e:
+                raise RuntimeError("AWS credentials are missing or incomplete. Please configure them.") from e
```
akotyla, line 52:
```suggestion
            except (NoCredentialsError, PartialCredentialsError, ClientError) as e:
                raise RuntimeError("AWS credentials are missing or incomplete. Please configure them.") from e
```

packages/ragbits-document-search/src/ragbits/document_search/documents/aws_s3_source.py
```diff
@@ -0,0 +1,183 @@
...
+    @classmethod
+    @requires_dependencies(["boto3"], "s3")
+    def _set_client(cls, bucket_name: str) -> None:
+        """
+        Set the boto3 S3 client if it hasn't been initialized yet.
+
+        Arguments:
+            bucket_name: The name of the S3 bucket to use.
+
+        Raises:
```
akotyla, line 40: `ClientError` is missing here

packages/ragbits-document-search/src/ragbits/document_search/documents/aws_s3_source.py
```diff
@@ -0,0 +1,183 @@
...
+
+        local_dir = get_local_storage_dir()
+        container_local_dir = local_dir / self.bucket_name
+        container_local_dir.mkdir(parents=True, exist_ok=True)
+        normalized_key = self.key.replace("/", "_")
+        path = container_local_dir / normalized_key
+
+        try:
+            self._s3_client.download_file(self.bucket_name, self.key, path)
+        except ClientError as e:
```
akotyla, line 82: I don't understand why we don't just raise exceptions here as printing errors here can cause silent failures, for example fetch will return a path even if the file doesn’t exist :(

packages/ragbits/pyproject.toml
```diff
@@ -46,6 +46,9 @@ chroma = [
 huggingface = [
     "datasets~=3.0.1",
 ]
+s3 = [
+    "boto3>=1.35.42",
```
micpst, line 50: just wondering, can we use `ragbits-document-search[s3]==0.8.0` here? defining the same requirements in two places doesn't seem right to me

akotyla:
I think we have a convention to name PRs like this: `<type>[optional scope]: <description>`, so `:` got lost here :)

packages/ragbits-document-search/pyproject.toml
```diff
@@ -50,7 +50,9 @@ huggingface = [
 distributed = [
     "ray>=2.39.0",
 ]
-
+s3 = [
+    "boto3>=1.35.42",
```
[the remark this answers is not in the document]
mhordynski: We are library, we should not be so strict about versions. It makes it hard to install ragbits later

packages/ragbits/pyproject.toml
mhordynski: it is done automatically in release process. All extras are copied to pyproject.toml in ragbits index package

packages/ragbits-document-search/pyproject.toml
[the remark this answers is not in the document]
micpst: i disagree we had problems in a past with incompatible versions of some libs like litellm that broke apis in later releases.

> it makes it hard to install ragbits later

i don't think so, quite the opposite, it should streamline the installation process, because we don't risk the user installing conflicting versions of different packages

packages/ragbits-document-search/src/ragbits/document_search/documents/aws_s3_source.py
kzamlynska: I added to the doc string

mhordynski (approved):

PR #354, opened 2025-02-11
feat: add history persistence component

akotyla:


packages/ragbits-conversations/src/ragbits/conversations/history/models.py
```diff
@@ -0,0 +1,52 @@
+import uuid
+
+from sqlalchemy import TIMESTAMP, Column, ForeignKey, Integer, String, func
+from sqlalchemy.orm import DeclarativeBase
+
+
+class Base(DeclarativeBase):
```
ludwiktrammer, line 7: If I understand correctly this is an internal detail connected to how SQLalchemy models work. If so, I would make the class private (`_Base`). You could also use the simplified syntax for declaring this class that is offered by SqlAlchemy (`_Base = declarative_base()`).

packages/ragbits-conversations/src/ragbits/conversations/history/store.py
```diff
@@ -0,0 +1,110 @@
...
+import json
+
+import sqlalchemy
+
+from ragbits.core.prompt import ChatFormat
+
+from .models import Conversation, Message
+
+
+class ConversationHistoryStore:
```
ludwiktrammer, line 11: I would expect that this SqlAlchemy-based implementation is just one of many possible implementations of history stores. Later we will most likely want to add other backends than just an SQL database.

Which means that I would:
1. Create an abstract base `ConversationHistoryStore` class that provides the interface for all history stores, and is an `ConfigurableComponent`.
2. Rename this implementation as to indicate that this is the SQL variant, and make sure it inherits from the base `ConversationHistoryStore`
3. Move all SQL_specific code to an `sql/` subdirectory

packages/ragbits-conversations/src/ragbits/conversations/history/store.py
```diff
@@ -0,0 +1,110 @@
...
+        """
+        Initializes the ConversationHistoryStore with a SQLAlchemy engine.
+
+        Args:
+            sqlalchemy_engine: The SQLAlchemy engine used to interact with the database.
+        """
+        self.sqlalchemy_engine = sqlalchemy_engine
+
+    @staticmethod
+    def generate_conversation_id(messages: ChatFormat) -> str:
```
ludwiktrammer, line 30: Why not use actual database ids? We can have multiple conversation with the same messages (especially the beginning of any conversation is often very similar) so updating conversations based just the content may lead to conversations of other users being accidentally updated.

ludwiktrammer (changes requested):

packages/ragbits-conversations/src/ragbits/conversations/history/models.py
akotyla: I changed it to a private class, but I had a problem with this simplified syntax you wrote about, `mypy` didn't work with it, and from what they write here: https://docs.sqlalchemy.org/en/20/orm/extensions/mypy.html, mypy plugin for sqlalchemy is deprecated and `DeclarativeBase` should be used: https://docs.sqlalchemy.org/en/20/changelog/whatsnew_20.html#step-one-orm-declarative-base-is-superseded-by-orm-declarativebase

packages/ragbits-conversations/src/ragbits/conversations/history/stores/base.py
```diff
@@ -0,0 +1,47 @@
+from abc import ABC, abstractmethod
+
+from ragbits.core.prompt import ChatFormat
+from ragbits.core.utils.config_handling import ConfigurableComponent
+
+
+class BaseHistoryStore(ConfigurableComponent, ABC):
```
ludwiktrammer, line 7: `ConfigurableComponent` requires a couple of static properties to be set on the base class: `options_cls`, `default_module`, `configuration_key` - see other `ConfigurableComponent`s for examples.

packages/ragbits-conversations/src/ragbits/conversations/history/store.py
akotyla: Fixed, thanks for such a comprehensive explanation :)

packages/ragbits-conversations/src/ragbits/conversations/history/store.py
akotyla: Ok, fixed

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
```diff
@@ -0,0 +1,140 @@
...
+class SQLHistoryStore(BaseHistoryStore):
+    """
+    A class to manage storing, retrieving, and updating conversation histories.
+
+    This class provides methods to create a new conversation, fetch an existing conversation,
+    and update a conversation with new messages. The conversations are stored in a SQLAlchemy-backed
+    database, and a unique conversation ID is generated based on the message contents.
+    """
+
+    def __init__(self, sqlalchemy_engine: sqlalchemy.Engine):
```
micpst, line 66: please use `AsyncEngine` here

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
```diff
@@ -0,0 +1,140 @@
...
+        Creates a new conversation in the database with an auto-generated ID.
+
+        Args:
+            messages: A list of dictionaries, where each dictionary represents a message
+            with 'role' and 'content' keys.
+
+        Returns:
+            The database-generated ID of the conversation.
+        """
+        with self.sqlalchemy_engine.connect() as connection:
```
micpst, line 86: if you're using orm, you should use session instead of connection

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Fetches a conversation by its ID.
+
+        Args:
+            conversation_id: The ID of the conversation to fetch.
+
+        Returns:
+            A list of message dictionaries, each containing 'role' and 'content'.
+        """
+        with self.sqlalchemy_engine.connect() as connection:
```
micpst, line 112: ditto

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
```diff
@@ -0,0 +1,140 @@
...
+        Updates a conversation with new messages.
+
+        Args:
+            conversation_id: The ID of the conversation to update.
+            new_messages: A list of new message objects in the chat format.
+
+        Returns:
+            The ID of the updated conversation.
+        """
+        with self.sqlalchemy_engine.connect() as connection:
```
micpst, line 129: ditto

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Attributes:
+        id: The unique identifier for the conversation.
+        created_at: The timestamp when the conversation was created.
+
+    Table:
+        conversations: Stores conversation records.
+    """
+
+    __tablename__ = "conversations"
```
micpst, line 27: wondering if we should prefix table names created by ragbits, it can be quite confusing if we integrate with an existing database and get new tables out of nowhere
cc @mhordynski

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
akotyla: why? in `db-ally` we didn't want to use `AsyncEngine` as I remember

mhordynski: As we offer an asynchronous public API, we should prefer the async versions of libraries wherever possible in ragbits. Synchronous calls to the database are blocking and can affect overall performance.

akotyla: ok, thanks for an explanation, changed

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
akotyla: done

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
akotyla: done

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
akotyla: fixed

packages/ragbits-conversations/src/ragbits/conversations/history/stores/base.py
akotyla: thanks, fixed :)

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        Initializes the ConversationHistoryStore with a SQLAlchemy engine.
+
+        Args:
+            sqlalchemy_engine: The SQLAlchemy engine used to interact with the database.
+            default_options: An optional SQLHistoryStoreOptions specifying table names.
+        """
+        super().__init__(default_options=default_options)
+        self.sqlalchemy_engine = sqlalchemy_engine
+        self.session = async_sessionmaker(sqlalchemy_engine, expire_on_commit=False)
```
micpst, line 95: these syncs for long running sessions can be slow, it would be better to create a new session inside each call (`create_conversation`, `fetch_conversation`, `update_conversation`) with `async with AsyncSession(engine) as session: ...`

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
micpst: i see you added configs, but default table names are still without prefix, can you fix this?

micpst: something like `ragbits_conversations` and `ragbits_messages` should be enough

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
akotyla: done

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
akotyla: done

packages/ragbits-conversations/src/ragbits/conversations/history/stores/sql.py
micpst: we can remove then this `self.session` ;)

akotyla: right, I'm blind...

akotyla: fixed :D

micpst (approved):

packages/ragbits-conversations/src/ragbits/conversations/history/stores/base.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class HistoryStore(ConfigurableComponent[HistoryStoreOptionsT], ABC):
+    """
+    Abstract base class for conversation history stores.
+    """
+
+    options_cls: type[HistoryStoreOptionsT]
+    default_module: ClassVar = stores
+    configuration_key: ClassVar = "store"
```
ludwiktrammer, line 19: We have different kinds of "stores" in the system, including vector stores, metadata stores. I don't think it would be intuitive that "history stores" are the ones that use the "store" key. I would change it to "history_store".

akotyla: Good point, fixed

ludwiktrammer (approved):

PR #360, opened 2025-02-12
fix: use common WhereQuery type in qdrant vector store 

kzamlynska:


packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
```diff
@@ -165,10 +178,28 @@ async def remove(self, ids: list[str]) -> None:
...
+        """
+        return Filter(
+            must=[
+                FieldCondition(key=key, match=MatchValue(value=typing.cast(bool | int | str, value)))
+                for key, value in where.items()
+            ]
+        )
+
     @traceable
     async def list(  # type: ignore
```
mhordynski, line 200: let's remove type ignore

mhordynski:
Please add entry to changelog.md

packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
kzamlynska: done that

micpst (approved):

packages/ragbits-core/src/ragbits/core/vector_stores/qdrant.py
```diff
@@ -189,12 +190,16 @@ def _create_qdrant_filter(where: WhereQuery) -> Filter:
...
-        )
+        try:
+            return Filter(
+                must=[
+                    FieldCondition(key=key, match=MatchValue(value=typing.cast(str | int | bool, value)))
+                    for key, value in where.items()
+                ]
+            )
+        except ValidationError:
+            print("ERROR: WhereQuery dictionary must contain only str | int | bool values")
```
micpst, line 201: i dont get this change, this print seems unnecessary

kzamlynska: Despite of the fact that the argument is WhereQuery which can be a dict with str, int, bool values in other vector stores* any dict can be used like nested one: {"a": {"b" : 2}}. in qdrant it can't as the qdrant filter can only take bool, str, int and float as a value.  I wanted to make it clear why that simple dict can b use for query.

*By other vector stores I mean pgvector and in_memory. In chroma I am not sure it works that way but it looks like it,

micpst: ok, but this print will be lost in a trace, maybe reraise this `ValidationError` with `ValueError` and custom message like this:

```python
try: ...
except ValidationError as exc:
    raise ValueError("ERROR: WhereQuery ...") from exc
```

ludwiktrammer: I agree with @micpst that this print doesn't make much sense - you should either report problems as exceptions or as warnings (if the problem is recoverable and the flow of the function can continue). Libraries shouldn't generally print anything (outside of CLI commands).

But also the message is not necessarily true - you have no way of knowing that the `ValidationError` has been thrown due to a wrong type (and not, for example, a wrong field name, or some other problem). If you want to specifically report wrong input types - check them explicitly yourself, and throw an exception.

kzamlynska: I've change the code and I added possibility of using nested dictionaries. the filtering will work for sure for dictionaries, will it work for list should be checked.
I remove the raising error.

mhordynski (approved):

PR #364, opened 2025-02-19
docs: how-to on using LLMs

akotyla:


docs/how-to/core/use_llms.md
```diff
@@ -0,0 +1,163 @@
...
+    llm = LiteLLM(model_name="gpt-4o-2024-08-06")
+    response = await llm.generate("Tell me a joke.")
+    print(response)
+
+asyncio.run(main())
+```
+
+With LiteLLM, you can switch between different LLM providers by changing the `model_name` parameter and configuring authentication accordingly. See the [LiteLLM documentation](https://docs.litellm.ai/docs/providers) for details on calling models from different providers.
+
+## Setting up and using a local models
```
mhordynski, line 23: I think we should move Local LLMs to its own document and here just link that. Right now it adds complexity to the guide and remote LLMs are our preferred approach.

akotyla: Good point, fixed

mhordynski (approved):
lgtm

kzamlynska (approved):
You added a lot of examples. I like it :)

PR #374, opened 2025-02-24
feat: add fastembed embeddings

mhordynski:


packages/ragbits-core/src/ragbits/core/embeddings/sparse.py
```diff
@@ -13,22 +13,18 @@
...
 
-@dataclass
-class SparseVector:
+class SparseVector(BaseModel):
     """Sparse Vector representation"""
 
-    non_zero_dims: list[int]
-    non_zero_vals: list[int]
-    dim: int
+    indices: list[int]
```
kdziedzic68, line 19: why have you removed info about dimension ? this is important in vector arithmetics: eg. losing this information we cannot compute vector's norm (as this is n-th root of sum of value squares)

packages/ragbits-core/src/ragbits/core/embeddings/fastembed.py
```diff
@@ -0,0 +1,77 @@
...
+
+        Returns:
+            List of embeddings for the given strings.
+        """
+        merged_options = (self.default_options | options) if options else self.default_options
+
+        return [[float(x) for x in result] for result in self._model.embed(data, **merged_options.dict())]
+
+
+class FastEmbedSparseEmbeddings(SparseEmbeddings[FastEmbedOptions]):
```
kdziedzic68, line 47: maybe as we hold separate sub-package for sparse base we should also split for the given provider ? or move base sparse to embeddings.base too

mhordynski: I think having both dense and sparse in one file for a provider is fine, but yeah moving base sparse may be a good idea. I'll create new ticket to refactor module structure and imports.

kdziedzic68 (approved):

PR #378, opened 2025-02-25
feat: audit tracing improvement

### LINKED ISSUE #344: feat: audit tracing improvements
### Feature description

The goal of this task is to refine our tracing setup:

1. Review the main pathways in our library, identify areas that are not currently traced, and add tracing to those sections.
2. Sometimes we send traces that should not be included in the tracing event, such as vectors. We need to eliminate these from the traces.
3. In the CLI handler, we currently truncate inputs/outputs that are too long. We should approach this more intelligently. For example, the full prompt should always be visible and formatted in a multiline style. Consider adding it as a rich output with `Syntax('markdown'/'json')`.

### Motivation

Tracing is one of core features of ragbits and should be improved continuously.

### Additional context

_No response_

kzamlynska:
closes #344

### DIFF
--- a/packages/ragbits-core/src/ragbits/core/audit/base.py
+++ b/packages/ragbits-core/src/ragbits/core/audit/base.py
@@ -3,7 +3,7 @@ from collections.abc import Iterator
 from contextlib import contextmanager
 from contextvars import ContextVar
 from types import SimpleNamespace
-from typing import Any, Generic, TypeVar
+from typing import Any, Generic, TypeVar, cast
 
 SpanT = TypeVar("SpanT")
 
@@ -88,7 +88,10 @@ class TraceHandler(Generic[SpanT], ABC):
         self.stop(outputs=vars(outputs), current_span=span)
 
 
-def format_attributes(data: dict, prefix: str | None = None) -> dict:
+# TODO: check do we need this method.
+
+
+def format_attributes_1(data: dict, prefix: str | None = None) -> dict:
     """
     Format attributes for CLI.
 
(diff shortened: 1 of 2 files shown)


packages/ragbits-core/src/ragbits/core/audit/base.py
```diff
@@ -88,7 +88,10 @@ def trace(self, name: str, **inputs: Any) -> Iterator[SimpleNamespace]:  # noqa:
         self.stop(outputs=vars(outputs), current_span=span)
 
 
-def format_attributes(data: dict, prefix: str | None = None) -> dict:
+# TODO: check do we need this method.
```
ludwiktrammer, line 91: What does this mean? Check how?

packages/ragbits-core/src/ragbits/core/audit/base.py
```diff
@@ -119,3 +122,112 @@ def format_attributes(data: dict, prefix: str | None = None) -> dict:
...
+    Returns:
+        The formatted attributes.
+    """
+    max_string_length = 150
+    max_list_length = 15
+    opt_list_lenght = 3
+
+    def shorten_list(lst: list | tuple) -> list:
+        """
+        Shortens a list if it's longer than 3 elements. Shortens list elements if it's long string.
```
ludwiktrammer, line 142:
```suggestion
        Shortens a list if it's longer than opt_list_lenght elements. Shortens list elements if it's long string.
```

packages/ragbits-core/src/ragbits/core/audit/base.py
```diff
@@ -119,3 +122,112 @@ def format_attributes(data: dict, prefix: str | None = None) -> dict:
             flattened[current_key] = repr(value)
 
     return flattened
+
+
+def format_attributes(data: dict, prefix: str | None = None) -> dict:
```
ludwiktrammer, line 127: This function with nested sub-functions that use local variables from the parent function pretty much looks like simulating a class with methods and instance properties. Maybe simply use a class if a class-like functionality is needed?

packages/ragbits-core/src/ragbits/core/audit/base.py
kzamlynska: changed

packages/ragbits-core/src/ragbits/core/audit/base.py
kzamlynska: I've changed the function to the class

packages/ragbits-core/src/ragbits/core/audit/base.py
kzamlynska: I've checked it. I needed to run the otel tracer and see how the new formater looks like.
I removed this method.

kzamlynska: After discussion I restored the format_attributes the way it was for otel trace handler .

kzamlynska:
I've added @traceable decorator to the most of the methods in the ragbits.core 
I've added @traceable decorator to some of the methods in other ragbits packages - (I've chosen the one with more complicated logic)
I've omitted ragbits.core.audit i ragbits.core.utils -> please let me know if we need tracing in the second one.
I've ommitted ragbits.cli

micpst (commented):
I think the main problems of the #344 issue are not solved in this PR. I'm talking about:

> 
> 1. Review the main pathways in our library, identify areas that are not currently traced, and add tracing to those sections.
> 
> 2. Sometimes we send traces that should not be included in the tracing event, such as vectors. We need to eliminate these from the traces.
> 

The main goal was to replace `traceable` decorators with `track` contexts with filtered arguments that are actually meaningful to the user, such as removing vector values, etc.

I don't see it in this PR, and I also feel that too much tracking has been added, such as tracking in private methods, which are irrelevant to the user.

kzamlynska:
If you want to see the newest prompt/response parsing please use examples/core/llm.py

mhordynski (changes requested):

mhordynski (approved):

PR #386, opened 2025-03-04
feat(core): Support images in few shot examples

konrad-czarnota-ds:


packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -94,31 +93,9 @@ def count_tokens(self, prompt: BasePrompt) -> int:
...
-                    },
-                }
-                for im in images
-            ]
-            last_message = chat[-1]
-            if last_message["role"] == "user":
-                user_message_content = [{"type": "text", "text": last_message["content"]}] + user_message_content
-                chat = chat[:-1]
-            chat.append({"role": "user", "content": user_message_content})
+        if prompt.has_images() and not litellm.supports_vision(self.model_name):
```
kdziedzic68, line 97: we are not doing anymore any kind of formatting - would rename the method to `_validate_chat_for_llm` or sth like that

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -252,9 +252,10 @@ async def generate_streaming(
             yield text_piece
 
     def _format_chat_for_llm(self, prompt: BasePrompt) -> ChatFormat:
-        if prompt.list_images():
-            wrngs.warn(message=f"Image input not implemented for {self.__class__.__name__}")
-        return prompt.chat
+        chat = prompt.chat
```
ludwiktrammer, line 255: nit: this temporary var doesn't seem necessary

packages/ragbits-core/src/ragbits/core/llms/litellm.py
ludwiktrammer: or simply move the check to the `_call` method

packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -37,14 +37,11 @@ def output_schema(self) -> dict | type[BaseModel] | None:  # noqa: PLR6301
         """
         return None
 
-    # ruff: noqa
-    def list_images(self) -> list[bytes | str]:
+    def has_images(self) -> bool:  # noqa: PLR6301
```
ludwiktrammer, line 40: Hm, I liked being able to list images by themselves - just as you can get the input and output messages from the prompt. I think it's a useful part of the prompt interface. It would be hard to, for example, add image support to Prompt Studio without being able to directly list images from a prompt.

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -198,13 +213,30 @@ def list_few_shots(self) -> ChatFormat:
...
+            for content in message["content"]
+            if isinstance(message["content"], list)
+        )
+
+    @staticmethod
+    def _create_message_with_image(image: str | bytes) -> dict:
+        return {
+            "type": "image_url",
+            "image_url": {
+                "url": f"data:image/jpeg;base64,{base64.b64encode(image).decode('utf-8')}"
```
ludwiktrammer, line 235: Sidenote: you didn't change this, so this is not something for this PR, but it seems weird to me that we assume all images to be jpegs.

packages/ragbits-core/src/ragbits/core/llms/litellm.py
konrad-czarnota-ds: moved

packages/ragbits-core/src/ragbits/core/prompt/base.py
konrad-czarnota-ds: Reverted - probably to be discussed in what format images should be returned, but I decided to simply extract the url from chat which is either base64 or a link - the way openAi support this

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
konrad-czarnota-ds: It turned out to be easy to fix, so I did it

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -145,7 +116,11 @@ async def _call(
             LLMConnectionError: If there is a connection error with the LLM API.
             LLMStatusError: If the LLM API returns an error status code.
             LLMResponseError: If the LLM API response is invalid.
+            LLMNotSupportingImagesError: If the model does not support images.
         """
+        if prompt.list_images() and not litellm.supports_vision(self.model_name):
+            raise LLMNotSupportingImagesError()
```
kdziedzic68, line 122: i would include the model name in the error message

konrad-czarnota-ds: added

micpst (commented):
Looks ok, let's maybe add new examples how to define few shots for prompts, one for text and one for multimodal few shots, we can also add example how to create multimodal prompt.

micpst (approved):

PR #387, opened 2025-03-04
docs: how-to on using images with prompts

akotyla:


docs/how-to/core/use_images_in_prompts.md
```diff
@@ -0,0 +1,124 @@
...
+from pydantic import BaseModel
+from ragbits.core.prompt import Prompt
+from ragbits.core.llms.litellm import LiteLLM
+
+
+class ImagePromptInput(BaseModel):
+    """
+    Input model containing a single image.
+    """
+    image: bytes | str
```
ludwiktrammer, line 22: nit: maybe it would be better to use more specific example with field names different then `image` / `images`, so it is more clear that this is not some required name?

docs/how-to/core/use_images_in_prompts.md
```diff
@@ -0,0 +1,124 @@
...
+    response = await llm.generate(prompt)
+    print(response)
+
+
+asyncio.run(main())
+```
+
+### Using multiple images as input
+
+If you need a prompt that accepts multiple images, define an input model containing a list of image strings.
```
ludwiktrammer, line 46: "image strings" is not really true - bytes or URLs

akotyla: fixed

PR #390, opened 2025-03-05
feat: evaluate testing

kdziedzic68:


packages/ragbits-evaluate/src/ragbits/evaluate/pipelines/base.py
```diff
@@ -39,3 +53,131 @@ async def prepare(self) -> None:
...
+    async def run_evaluation(
+        self,
+        dataloader: DataLoader,
+        metrics: MetricSet,
+    ) -> dict:
+        """
+        Compute the evaluation results for the given pipeline and data.
+
+        Args:
+            pipeline: The pipeline to be evaluated.
```
konrad-czarnota-ds, line 87: I don't see an argument pipeline

packages/ragbits-evaluate/src/ragbits/evaluate/pipelines/base.py
```diff
@@ -1,24 +1,38 @@
...
+    pipeline: ObjectContructionConfig
+    metrics: dict[str, ObjectContructionConfig]
+
 
 class EvaluationPipeline(Generic[EvaluationTargetT], WithConstructionConfig, ABC):
     """
     Collection evaluation pipeline.
     """
 
+    CONCURRENCY: int = 10
```
konrad-czarnota-ds, line 34: Wouldn't it be better to have this as a part of EvaluationConfig so it could be changed if needed?

packages/ragbits-evaluate/src/ragbits/evaluate/pipelines/base.py
```diff
@@ -39,3 +53,131 @@ async def prepare(self) -> None:
...
+
+        Args:
+            config: Evaluation config.
+
+        Returns:
+            The evaluation results.
+        """
+        model = EvaluationConfig.model_validate(config)
+        dataloader: DataLoader = DataLoader.subclass_from_config(model.dataloader)
+        pipeline: EvaluationPipeline = EvaluationPipeline.subclass_from_config(model.pipeline)
```
micpst, line 70: After checking the code, I am no longer sure that moving `Evaluator` to `EvaluationPipeline` is a good move. Here you are creating an `EvaluationPipeline` object inside `EvaluationPipeline`. This is confusing for the user and mixes two different abstractions that imho should be separated - the code being evaluated and the evaluation orchestration. I think we should undo this change and bring back `Evaluator`.

micpst (changes requested):

packages/ragbits-evaluate/src/ragbits/evaluate/pipelines/base.py
kdziedzic68: changed it to be a class method now. would not say it is confusing - as well one can say that all class methods are confusing. To me stateless object is more confusing than this and actually the code which is being evaluated is the object `evaluation_target` (eg. DocumentSearch) and `EvaluationPipeline` is already kind of orchestration for eval execution

packages/ragbits-evaluate/src/ragbits/evaluate/pipelines/base.py
kdziedzic68: added

packages/ragbits-evaluate/src/ragbits/evaluate/pipelines/base.py
kdziedzic68: good point. also renamed the parameter to `batch_size` - should be more intuitive. probably we should also support multiple batching strategies here as we do withing document processing

micpst (approved):

PR #399, opened 2025-03-11
fix: better support for mixed text/image embeddings in Qdrand and Chroma

ludwiktrammer:


packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
```diff
@@ -119,24 +105,26 @@ async def store(self, entries: list[VectorStoreEntry]) -> None:
...
-            metadatas.append(
-                self._flatten_metadata(
-                    {
-                        **entry.metadata,
-                        **{
-                            "__embeddings": raw_embeddings[entry.id],
-                            "__image": entry.image_bytes.hex() if entry.image_bytes else None,
-                        },
-                    }
+            for embeddint_type in self._embedding_types:
```
mhordynski, line 109: embeddint -> embedding

kzamlynska:
I found only one problem, chroma example is not working:
examples/document-search/chroma.py
I think it is related to the previous refactoring, but maybe it is good idea to correct it here?

ludwiktrammer:
@kzamlynska Thanks! I believe It currently works on `main` and I broke it in this PR. I will fix it.

ludwiktrammer:
@kzamlynska Thanks, fixed!

mhordynski (approved):

PR #404, opened 2025-03-12
feat: add HttpSource

mackurzawa:
Description
* Added HttpSource, so that users can use it as the source by providing a URL to the file
* Added tests for HttpSource


packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+    import validators
+
+from ragbits.core.utils.decorators import requires_dependencies
+from ragbits.document_search.documents.sources import Source
+from ragbits.document_search.documents.sources.base import get_local_storage_dir
+
+HTTP_STATUS_OK = 200
+
+
+class HttpSource(Source):
```
micpst, line 17: Please add `protocol` attribute, otherwise `SourceResolver` won't work for this source

micpst: The `list_sources` function should also be implemented

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+        container_local_dir.mkdir(parents=True, exist_ok=True)
+        path = container_local_dir / normalized_url_path
+
+        async with aiohttp.ClientSession() as session, session.get(self.url) as response:
+            if response.status == HTTP_STATUS_OK:
+                with open(path, "wb") as f:
+                    async for chunk in response.content.iter_chunked(1024):
+                        f.write(chunk)
+            else:
+                raise ValueError(f"Failed to download {self.url}. Response code: {response.status}")
```
micpst, line 57: We have some custom source exceptions defined in `ragbits.document_search.documents.exceptions`, you can use them here or add new ones

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Returns:
+            A sequence containing a HttpSource instance.
+
+        Raises:
+            ValueError: If the url has invalid format
+
+        """
+        if not validators.url(url):
+            raise ValueError(f"Invalid URL: {url}")
```
micpst, line 80: ditto

packages/ragbits-document-search/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Compability with the new Vector Store interface from ragbits-core (#288)
 - Fix docstring formatting to resolve Griffe warnings
+- Add HttpSource (#397)
```
ludwiktrammer, line 7: This is not particularly meaningful for users reading the changelog - instead of just mentioning the name of the class, I would instead write what it does.

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+
+with suppress(ImportError):
+    import aiohttp
+    import validators
+
+from ragbits.core.utils.decorators import requires_dependencies
+from ragbits.document_search.documents.sources import Source
+from ragbits.document_search.documents.sources.base import get_local_storage_dir
+
+HTTP_STATUS_OK = 200
```
ludwiktrammer, line 14: this is already built into Python standard library - `HTTPStatus.OK` from `http`, but when working with responses from the aiohttp library you can simply do `if response.ok:` instead of comparing the status.

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+
+class HttpSource(Source):
+    """
+    An object representing an HTTP dataset source.
+    """
+
+    url: str
+
+    @property
+    def id(self) -> str:
```
ludwiktrammer, line 25: For other source types we prepend the ids with some sort of an internal protocol, so that they won't collide between sources. I think it would be worth doing the same here.

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+
+        Returns:
+            Path: The local path to the downloaded file.
+
+        Raises:
+            ValueError: If the download fails.
+        """
+        parsed_url = urlparse(self.url)
+        domain_name = parsed_url.netloc
+        normalized_url_path = parsed_url.path.replace("/", "_")
```
ludwiktrammer, line 44: I think there might be more problematic characters than just "/". I would simply replace all `\W` characters.

packages/ragbits-document-search/pyproject.toml
```diff
@@ -58,6 +58,10 @@ distributed = [
 s3 = [
     "boto3~=1.35.42",
 ]
+http = [
```
mhordynski, line 61: let's move it to default dependencies, this is very common so I think we should be fine with that

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,82 @@
...
+            url: The URI path.
+
+        Returns:
+            A sequence containing a HttpSource instance.
+
+        Raises:
+            ValueError: If the url has invalid format
+
+        """
+        if not validators.url(url):
```
ludwiktrammer, line 79: I wouldn't introduce a new library just for this - you can use `urlparse` from `urllib.parse`. To be honest, I also don't believe it is the responsibility of the `from_uri()` method to do more validation that is done when creating the object normally.

ludwiktrammer (changes requested):

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
mackurzawa: True! Both added

packages/ragbits-document-search/pyproject.toml
mackurzawa: Moved to default dependencies

packages/ragbits-document-search/CHANGELOG.md
mackurzawa: Changed!

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
mackurzawa: Great tip! Implemented

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
mackurzawa: Done

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
kzamlynska: > The `list_sources` function should also be implemented
In fact the list_sources is not defined in the Source class, so it is not required. It is used only when we do from_uri and the target of uri is a folder/container/blob, not a singular file. Then we list_sources.
What could be the use case of list sources from http?

packages/ragbits-document-search/tests/unit/test_http_source.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_id():
+    source = HttpSource(url="http://example.com/file.pdf")
+    expected_id = "http://example.com/file.pdf"
+    assert source.id == expected_id
+
+
+async def test_from_uri_one_file():
+    file_uris = [
```
kzamlynska, line 14: The uri "https://google.com" - is a good uri according to your parsing method but doesn't point the singular file.

ludwiktrammer: @kzamlynska It can. There is nothing in the HTTP protocol that says that a file cannot be located under the root of a domain.

As an example: sometime ago, for a security-related demonstration,  I bouth a `documents-backup.zip` domain and put a .zip file directly in the root of the domain:

http://documents-backup.zip

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
mackurzawa: Added!

mhordynski (approved):
lgtm

mhordynski:
Let's add HttpSource to `packages/ragbits-document-search/src/ragbits/document_search/documents/sources/__init__.py`

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/http.py
```diff
@@ -0,0 +1,92 @@
...
+from ragbits.document_search.documents.sources.base import get_local_storage_dir
+
+
+class HttpSource(Source):
+    """
+    An object representing an HTTP dataset source.
+    """
+
+    url: str
+    protocol: ClassVar[str] = "web"
```
mhordynski, line 23:
```suggestion
    protocol: ClassVar[str] = "https"
```

mackurzawa: In that case, I assume it would be good to change it from `web` to `https` in the `id()` method too.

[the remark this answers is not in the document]
mackurzawa: Done

mackurzawa: Done

mhordynski (approved):

PR #417, opened 2025-03-17
feat: optional config_model in WithConstructionConfig

### LINKED ISSUE #237: feat: add (optional) config_model field for all WithConstructionConfig objects
### Feature description

Each object inheriting from `WithConstructionConfig` should have additional class variable: `config_model` (None in base class).

default implementation of `from_config` should:

* check if class has non-empty `config_model`
* if so validate the config with respect to it
* instantiate the object 

### Motivation

The process would simplify objects instantiation - the configs would be validated under the hood

### Additional context

_No response_

kzamlynska:
closes #237


packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -86,8 +86,11 @@ class WithConstructionConfig(abc.ABC):
     # The key under configuration for this class (and its subclasses) can be found.
     configuration_key: ClassVar[str]
 
+    # The type of pydantic model
+    model_type: ClassVar[type[BaseModel] | None] = None
```
ludwiktrammer, line 90: This will be used in a way in inheriting classes, in a way that will not make it immediately clear that this property is connected with `WithConstructionConfig` (for example, it will be seen as simply `model_type` property of the `VectorStore` class), so I would rename it to make it clear that this is a model for configuration. For example `config_model`, `config_validation`, `config_validation_model`

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -174,8 +177,22 @@ def from_config(cls, config: dict) -> Self:
...
+
+        Raises:
+            AttributeError: If model_type is not a subclass of BaseModel.
+            InvalidConfigError: If the config is invalid with respect to given model type.
         """
-        return cls(**config)
+        if cls.model_type:
+            try:
+                validated_config = cls.model_type.model_validate(config)
+                return cls(**validated_config.model_dump())
```
ludwiktrammer, line 188: Seems like this may cause loss of some information, for example when we use objects as values - which almost all classes that overwrite `from_config` do (see `QdrantVectorStore` as an example). Maybe validate but don't serialize only to immediately deserialize?

packages/ragbits-conversations/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## Unreleased
+- correct the typo in the class name ObjectConstructionConfig
```
ludwiktrammer, line 4: You should now be able to replace this with `Changelog-ignore: ragbits-conversations` in the commit message

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
kzamlynska: done

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
kzamlynska: renamed it to confgi_model

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
```diff
@@ -204,7 +221,21 @@ def from_config(cls, config: dict[str, Any]) -> ConfigurableComponent:
...
             An instance of the class initialized with the provided configuration.
+
+        Raises:
+            AttributeError: If model_type is not a subclass of BaseModel.
+            InvalidConfigError: If the config is invalid with respect to given model type.
         """
         default_options = config.pop("default_options", None)
         options = cls.options_cls(**default_options) if default_options else None
-        return cls(**config, default_options=options)
+        if cls.model_type:
```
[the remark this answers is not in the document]
kzamlynska: Please check if my correction is ok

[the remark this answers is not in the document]
kzamlynska: If I call super() there is a problem with lack of the default_options in WithConstructionConfig init.

packages/ragbits-core/src/ragbits/core/utils/config_handling.py
kzamlynska: I am suing model_dump because I need to return the class instance. And I don't know how to do it differently.

kzamlynska:
We decided there is no point to implement this feature now.

PR #422, opened 2025-03-18
fix: Rename HttpSource to WebSource

mackurzawa:
# Description
* Renamed `HttpSource` to `WebSource`
* Changed `url` to `uri` in field name, arguments, docstrings
* Changed prefix in `id()` to `web:`
* Introduced `WebDownloadError`

### DIFF
--- a/packages/ragbits-document-search/src/ragbits/document_search/documents/sources/web.py
+++ b/packages/ragbits-document-search/src/ragbits/document_search/documents/sources/web.py
@@ -9,41 +9,42 @@ with suppress(ImportError):
     import aiohttp
 
 from ragbits.core.utils.decorators import requires_dependencies
-from ragbits.document_search.documents.exceptions import SourceNotFoundError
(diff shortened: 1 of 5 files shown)


mhordynski (approved):

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/web.py
```diff
@@ -9,84 +9,85 @@
...
 
-class HttpSource(Source):
+class WebSource(Source):
     """
-    An object representing an HTTP dataset source.
+    An object representing a Web dataset source.
     """
 
-    url: str
+    uri: str
```
ludwiktrammer, line 22: Not that this is particularly important, but I wonder why? To me "URL" seems more appropriate in this case (i.e., needing a _location_ to fetch data from). "URI", on the other hand seem to broad since it may refer to URIs that do not specify locations - which we can't use here.

PR #425, opened 2025-03-19
refactor: rename properties and update docstrings in WebSource

mackurzawa:
# Description
* rename properties and update docstrings in WebSource

### DIFF
--- a/packages/ragbits-document-search/src/ragbits/document_search/documents/sources/web.py
+++ b/packages/ragbits-document-search/src/ragbits/document_search/documents/sources/web.py
@@ -65,17 +65,17 @@ class WebSource(Source):
         return path
 
     @classmethod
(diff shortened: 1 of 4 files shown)


packages/ragbits-document-search/src/ragbits/document_search/documents/sources/web.py
```diff
@@ -70,12 +70,12 @@ async def list_sources(cls, uri: str) -> Sequence["WebSource"]:
         List the file under the given URI.
 
         Arguments:
-            uri: The URI to the file.
+            uri: The URI to the file. Needs to include the protocol.
```
ludwiktrammer, line 73: This should be `url`. Unlike `from_uri`, in `list_sources` the types of arguments are source-specific (different for each source), and in case of `WebSource`, a proper URL is needed.

pawel-chmielak-deepsense (approved):

PR #427, opened 2025-03-19
feat: add Prometheus & Grafana Monitoring for LLMs Using OpenTelemetry

akotyla:


packages/ragbits-core/src/ragbits/core/audit/otel_metric_handler.py
```diff
@@ -0,0 +1,44 @@
...
+            ),
+            "input_tokens": self._meter.create_histogram(
+                name="input_tokens", description="Tracks the number of input tokens per request", unit="tokens"
+            ),
+            "time_to_first_token": self._meter.create_histogram(
+                name="time_to_first_token", description="Tracks the time to first token in milliseconds", unit="ms"
+            ),
+        }
+
+    def record(self, metric_name: str, value: float, attributes: dict | None = None) -> None:
```
mhordynski, line 32: I think we should have some enum for metrics, passing string directly is error-prone.

packages/ragbits-core/src/ragbits/core/audit/__init__.py
```diff
@@ -6,18 +6,56 @@
...
+_metric_handler: OtelMetricHandler | None = None
+
 
 Handler = str | TraceHandler
 
 P = ParamSpec("P")
 R = TypeVar("R")
 
 
+def set_metric_handler(meter: Meter) -> None:
```
mhordynski, line 34: It should be configurable to add some sort of prefix to all metrics. For example `input_tokens` -> `ragbits_input_tokens`

docs/how-to/project/use_metrics.md
```diff
@@ -0,0 +1,63 @@
...
+
+Each metric is tagged with the following **labels** (attributes) for better filtering and analysis: the class name of the prompt being processed (`promp`) and the name of the LLM being used (`model`).
+
+## Visualize Metrics in Grafana
+
+To visualize the metrics collected by Ragbits, follow these steps:
+
+1. Open your browser and navigate to http://localhost:3000.
+2. Use the default credentials to log in (username: `admin`, password: `admin`).
+3. Once logged in, you can either navigate to the Explore section to query and visualize metrics directly or create a new dashboard to build custom visualizations tailored to your needs.
```
mhordynski, line 63: I think we should provide some prebuilt dashboard :D

packages/ragbits-core/src/ragbits/core/audit/otel_metric_handler.py
akotyla: Thanks, fixed :)

docs/how-to/project/use_metrics.md
akotyla: I added an example dashboard in examples

examples/core/audit/grafana_dashboard.json
micpst, line 1: I don't really know what to do with this file :( 
Let's maybe write a simple script here where you call LLM with a prompt and enable traces and metrics in otel. We can keep this file, I assume it is important for visualization, but in the script write a docstring on how to use it. You can look at the examples in the document search on how to write docstring for examples. You can also refer to this script in the how-to.

micpst (approved):

PR #428, opened 2025-03-19
feat: add a hybrid vector store implementation

ludwiktrammer:


packages/ragbits-core/src/ragbits/core/vector_stores/hybrid.py
```diff
@@ -0,0 +1,163 @@
...
+        remove_tasks = (vector_store.remove(ids) for vector_store in self.vector_stores)
+        await asyncio.gather(*remove_tasks)
+
+    @traceable
+    async def list(
+        self, where: WhereQuery | None = None, limit: int | None = None, offset: int = 0
+    ) -> list[VectorStoreEntry]:
+        """
+        List entries from the vector stores. The entries can be filtered, limited and offset.
+        Vector wstores are queried in the order they were provided in the constructor.
```
kzamlynska, line 138:
```suggestion
        Vector stores are queried in the order they were provided in the constructor.
```

packages/ragbits-core/src/ragbits/core/vector_stores/hybrid.py
```diff
@@ -0,0 +1,163 @@
...
+                Not specifying the key means no filtering.
+            limit: The maximum number of entries to return.
+            offset: The number of entries to skip.
+
+        Returns:
+            The entries.
+        """
+        retrieved_results: dict[UUID, VectorStoreEntry] = {}
+        for vector_store in self.vector_stores:
+            needed = None
```
kzamlynska, line 151:
```suggestion
 
```

packages/ragbits-core/src/ragbits/core/vector_stores/hybrid.py
```diff
@@ -0,0 +1,163 @@
...
+
+        Args:
+            vector_stores: The vector stores to proxy requests to.
+            retrieval_strategy: The retrieval strategy to use when combining results,
+                uses OrderedHybridRetrivalStrategy by default.
+        """
+        self.vector_stores = vector_stores
+        self.retrieval_strategy = retrieval_strategy or OrderedHybridRetrivalStrategy()
+
+    @traceable
```
kzamlynska, line 84: If we use the 'with trace' in all of the methods in this class we could have in trace the info about vector_stores used. It can be valuable for debugging.

packages/ragbits-core/src/ragbits/core/vector_stores/base.py
```diff
@@ -147,53 +142,44 @@ class VectorStoreWithExternalEmbedder(VectorStore[VectorStoreOptionsT]):
     def __init__(
```
kzamlynska, line 142: please correct the typo in line 130:
 Types of embbedings supported by the vector store. ->  Types of embeddings supported by the vector store.

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -48,6 +47,7 @@ def __init__(
         table_name: str,
         vector_size: int,
         embedder: Embedder,
+        embedding_type: EmbeddingType = EmbeddingType.TEXT,
```
kzamlynska, line 50: Maybe we cab add embedding_type to with trace in store and retrieve. It can be handy sometimes.

packages/ragbits-core/src/ragbits/core/vector_stores/in_memory.py
```diff
@@ -25,6 +25,7 @@ class InMemoryVectorStore(VectorStoreWithExternalEmbedder[VectorStoreOptions]):
     def __init__(
         self,
         embedder: Embedder,
+        embedding_type: EmbeddingType = EmbeddingType.TEXT,
```
kzamlynska, line 28: Here we also can add embedding_type to the with trace of store and retrieve methods.

packages/ragbits-core/src/ragbits/core/vector_stores/chroma.py
```diff
@@ -44,12 +43,14 @@ def __init__(
...
             index_name: The name of the index.
             embedder: The embedder to use for converting entries to vectors.
+            embedding_type: Which part of the entry to embed, either text or image. The other part will be ignored.
             distance_method: The distance method to use.
             default_options: The default options for querying the vector store.
         """
         super().__init__(
             default_options=default_options,
             embedder=embedder,
+            embedding_type=embedding_type,
```
kzamlynska, line 53: And in Chromae we also can add embedding_type to the with trace of store and retrieve methods.

packages/ragbits-core/tests/integration/vector_stores/test_vector_store.py
```diff
@@ -20,37 +28,63 @@
 ]
 
 image_embbedings = [
-    [[0.7, 0.8, 0.9], [1.0, 0.81, 0.84]],  # for storage
```
kzamlynska, line 23: please rename image_embbedings to image_embeddings

packages/ragbits-core/tests/integration/vector_stores/test_vector_store.py
```diff
@@ -20,37 +28,63 @@
...
+    """
+    For each vector store in `vector_store_cls`, returns two instances of it, one for text and one for image embeddings.
+    """
+    embedder = (
+        NoopEmbedder(return_values=text_embbedings)
+        if request.param == EmbeddingType.TEXT
+        else NoopEmbedder(return_values=text_embbedings[-1:], image_return_values=image_embbedings)
+    )
+
+    # Workarond for Chroma reusing resources between EphemeralClient() instances
```
kzamlynska, line 69:
```suggestion
    # Workaround for Chroma reusing resources between EphemeralClient() instances
```

packages/ragbits-core/tests/integration/vector_stores/test_vector_store.py
```diff
@@ -162,7 +203,7 @@ async def test_handling_document_ingestion_with_different_content_and_verifying_
     document_2 = DocumentMeta.create_text_document_from_literal(document_2_content)
 
     document_search = DocumentSearch(
-        vector_store=vector_store,
+        vector_store=text_vector_store,
```
kzamlynska, line 206: My comment is regards the line 210: If not isinstance(document_2.source, LocalFileSource)  we don't have document_2_path. If we assume that it is (it is a test so it s) than we don't need the if statement here.

packages/ragbits-core/src/ragbits/core/vector_stores/hybrid.py
ludwiktrammer: Wouldn't the infor come automatically from those vector stores also sending their traces?

packages/ragbits-core/tests/integration/vector_stores/test_vector_store.py
ludwiktrammer: I haven't touch this test, but I'm pretty sure the author of the test used `isinstance` here to show the type checker that `document_2.source` is indeed an instance `LocalFileSource` (as the `source` property in documents is typed simply as `Source`). That's a much better method then adding `# type: ignore`.

packages/ragbits-core/src/ragbits/core/vector_stores/hybrid.py
kzamlynska: I'm pretty sure that traceable takes only explicit methods arguments. 

For example in case of store method it will be only entries, for list method only where, limit, offset.
When use with trace we can add more arguments to trace like self.vector_stores or self.retrieval_strategy>

If you think it is not necessary for the users I will not oppose. :)

ludwiktrammer: What I meant was that when you run methods on `HybridSearchVectorStore`, it will in turn run the same methods on its underlying vector stores - which will also send those to tracing.

So what you will end up with is a trace tree with individual stores being listed underneath `HybridSearchVectorStore`

kzamlynska: Ok :)

kzamlynska (approved):

mhordynski (approved):

PR #432, opened 2025-03-20
feat: response_parser both sync and async

### LINKED ISSUE #280: feat: allow for asynchronous parsers in Prompt
### Feature description

`Prompt` class should allow to pass asynchronous function as a parser. During parsing we should use asyncio.iscoroutine check to decide whether parser should be awaited or not.

### Motivation

In some cases user may want to perform asynchronous call while parsing LLM response.

### Additional context

_No response_

kzamlynska:
closes #280


packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -16,6 +17,9 @@
 InputT = TypeVar("InputT", bound=BaseModel | None)
 FewShotExample = tuple[str | InputT, str | OutputT]
 
+SyncParser = Callable[[str], OutputT]
+AsyncParser = Callable[[str], Awaitable[OutputT]]
```
ludwiktrammer, line 21: Where are those used?

kzamlynska: Nowhere. I removed them Thanks for notice

packages/ragbits-core/tests/unit/prompts/test_prompt.py
```diff
@@ -515,3 +516,26 @@ class TestPrompt(Prompt[_PromptInput, str]):  # pylint: disable=unused-variable
...
         {"role": "user", "content": "Theme for the song is rock."},
     ]
+
+
+async def test_response_parser():
+    class TestPrompt(Prompt):
+        user_prompt = "Hello AI"
+
+    async def async_parser(response: str) -> str:
+        await asyncio.sleep(1)
```
mhordynski, line 526: we don't need that sleep here - it will just make tests slower

mhordynski (approved):

PR #435, opened 2025-03-20
fix: Make source errors picklable

### LINKED ISSUE #434: bug: source errors are not picklable
### What happened?

Exceptions defined in `packages/ragbits-document-search/src/ragbits/document_search/documents/exceptions.py` are not picklable, which means they cannot be returned by processing done in Ray.

### How can we reproduce it?

```python
from ragbits.document_search.documents.exceptions import WebDownloadError
import pickle

err = WebDownloadError("123", "400")
err_pickled = pickle.dumps(err)
pickle.loads(err_pickled)
```

### Relevant log output

```shell
TypeError: WebDownloadError.__init__() missing 1 required positional argument: 'code'
```

pawel-chmielak-deepsense:
Fixes #434 by implementing `__reduce__` method in `SourceError`.

### DIFF
--- a/packages/ragbits-document-search/src/ragbits/document_search/documents/exceptions.py
+++ b/packages/ragbits-document-search/src/ragbits/document_search/documents/exceptions.py
@@ -1,3 +1,7 @@
+import inspect
+from typing import Any, Self
+
+
 class SourceError(Exception):
     """
     Class for all exceptions raised by the document source.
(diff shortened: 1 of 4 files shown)


mhordynski (approved):
lgtm

packages/ragbits-document-search/src/ragbits/document_search/documents/exceptions.py
```diff
@@ -7,6 +11,19 @@ def __init__(self, message: str) -> None:
         super().__init__(message)
         self.message = message
 
+    def __reduce__(self) -> tuple[type["SourceError"], tuple[Any, ...]]:
```
micpst, line 14: Why is it not pickable? it has only basic types

pawel-chmielak-deepsense: When it tries to recreate the exception it uses base class's \_\_reduce\_\_, which calls the correct class's initializer, but provides only `message` parameter (because that's what the base class uses).

Because of that any other parameter like `code` or `source_id` is not provided at all.

PR #447, opened 2025-03-25
chore: add integration tests to s3 source

mhordynski:
Changelog-ignore: ragbits-document-search


packages/ragbits-document-search/src/ragbits/document_search/documents/sources/s3.py
```diff
@@ -116,7 +116,7 @@ async def list_sources(cls, bucket_name: str, prefix: str) -> Sequence["S3Source
         cls._set_client(bucket_name)
         if cls._s3_client is None:
             raise RuntimeError("S3 client is not initialized.")
-        with trace(bucket=cls.bucket_name, key=cls.key) as outputs:
+        with trace(bucket=bucket_name, key=prefix) as outputs:
```
kzamlynska, line 119: Shouldn't we change this method and run it on cls.bucket_name and not have bucket_name as an argument?
We need the list method to list the files in one bucket : the one for which we created an instance of S2Source.

packages/ragbits-document-search/tests/integration/test_s3.py
```diff
@@ -0,0 +1,85 @@
+import os
+from collections.abc import Generator
+from pathlib import Path
+
+import boto3
+import pytest
+from moto import mock_s3
```
kzamlynska, line 7: Tests are passing. Everything is ok. I have only one question?
Shouldn;t we call that unit tests? My impression is that if we mock s3 it is not really integration test.

packages/ragbits-document-search/src/ragbits/document_search/documents/sources/s3.py
mhordynski: There is no such thing as cls.bucket_name, this is a classmethod so there is no instance of S3Source.

packages/ragbits-document-search/tests/integration/test_s3.py
mhordynski: yeah, this is kinda blend of unit and integration tests. As we're simulating real aws environment with moto and communicating with it one could say that it is an integration test, but it's not 100% accurate. 

I don't have any strong opinion about it, decided to have them here to separate it from test that are already there - as they have a little different nature.

kzamlynska (approved):

PR #458, opened 2025-03-27
feat: embed_image for liteLLM

### LINKED ISSUE #433: feat: add image support to `LiteLLMEmbedder`
`LiteLLMEmbedder` should support image embeddings when used with embedding models that support image embeddings.

We previously attempted to add this capability, but LiteLLM had issues that made the multimodal embedder API completely inconsistent with the API of other models. This made it exceedingly difficult to integrate support for these models into our existing LiteLLM client class. As a temporary workaround, we introduced `VertexAIMultimodelEmbedder`, which also uses LiteLLM behind the scenes but only supports a single family of multimodal models.

Given the current state of LiteLLM, it appears that the upstream issues have been resolved, making it relatively straightforward to add image support to our `LiteLLMEmbedder` class.

### Definition of Done

1. `LiteLLMEmbedder` should support embedding images via the `embed_image` method when configured with a model that supports it.  
2. The `image_support` method in `LiteLLMEmbedder` should correctly indicate whether the model supports image embeddings. (Note: In cases of uncertainty, false negatives are preferable to false positives.)  
3. The `VertexAIMultimodelEmbedder` class should be removed, and all instances of its use should be replaced by `LiteLLMEmbedder` configured with the `vertex_ai/multimodalembedding` model.

kzamlynska:
closes #433

### DIFF
--- a/packages/ragbits-core/CHANGELOG.md
+++ b/packages/ragbits-core/CHANGELOG.md
@@ -1,6 +1,7 @@
 # CHANGELOG
 
 ## Unreleased
+- image embeddings with liteLLM for vertex_ai models
 
 ## 0.12.0 (2025-03-25)
 - Allow Prompt class to accept the asynchronous response_parser. Change the signature of parse_response method.
--- a/packages/ragbits-core/src/ragbits/core/embeddings/litellm.py
+++ b/packages/ragbits-core/src/ragbits/core/embeddings/litellm.py
@@ -1,4 +1,9 @@
+import asyncio
+import base64
+
 import litellm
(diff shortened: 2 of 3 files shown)


kzamlynska:
There is still a problem with liteLLM library: 
- we can't distinguish between models supporting images and those one which not.
- some of the models are able to work with list of objects for embedding some not - liteLLM doesn't deal with this problem.
The solution which we have know with separate vertex_multimodal class is good at this moment and the possible improvements will make the code more complicated.
Decision: we won't work on this problem now.

PR #463, opened 2025-03-29
refactor: make litellm and litellmembedder consistent

mhordynski:


packages/ragbits-core/src/ragbits/core/embeddings/litellm.py
```diff
@@ -85,17 +87,17 @@ async def embed_text(self, data: list[str], options: LiteLLMEmbedderOptions | No
...
             options=merged_options.dict(),
         ) as outputs:
             try:
                 entrypoint = self.router or litellm
                 response = await entrypoint.aembedding(
                     input=data,
-                    model=self.model,
-                    api_base=self.api_base,
+                    model=self.model_name,
+                    base_url=self.base_url,
```
micpst, line 100: Are you sure this will work? afaik `litellm.aembedding` expects here `api_base` instead of `base_url`.

btw, maybe we should call it `api_base` everywhere, since it's tied to `api_version` and `api _key` in a way, `base url` is less informative.

packages/ragbits-core/src/ragbits/core/embeddings/vertex_multimodal.py
```diff
@@ -86,8 +87,8 @@ async def _embed(self, data: list[dict], options: LiteLLMEmbedderOptions | None
         merged_options = (self.default_options | options) if options else self.default_options
         with trace(
             data=data,
-            model=self.model,
-            api_base=self.api_base,
+            model=self.model_name,
+            base_url=self.base_url,
```
micpst, line 91: this needs to be changed too

micpst (approved):

PR #480, opened 2025-03-31
feat: add history to prompt

mhordynski:


packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -216,6 +220,57 @@ def list_few_shots(self) -> ChatFormat:
...
+
+        Args:
+            message (str): The assistant message content.
+
+        Returns:
+            Prompt[InputT, OutputT]: The current prompt instance to allow chaining.
+        """
+        if isinstance(message, BaseModel):
+            message = message.model_dump_json()
+        self._conversation_history.append({"role": "assistant", "content": message})
```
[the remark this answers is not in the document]
mhordynski: good point, fixed

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -166,6 +169,7 @@ def chat(self) -> ChatFormat:
             ),
             *self.list_few_shots(),
             {"role": "user", "content": user_content},
+            *self._conversation_history,
```
micpst, line 172: history should be above the user message

mhordynski: I think it should be after user message.

It works like this:

* we render system message, fewshots, user message while initializing the Prompt instance (for example `MyPrompt(MyPromptInput(...)`
* only after this we have access to instance methods `add_user/assistant_message`
* when we've added user or assistant message it means we did it after rendering the initial one, so it should be after

micpst: um, ok. I got a bit confused here. You're right, it'll work, but I'm not sure about the final api. Here is how this is going to work now:

```python
question = input()
prompt = KeenChatterPrompt(KeenChatterInput(tone="nice", question=question))

while True: 
    response = await llm.generate(prompt)
    prompt.add_assistant_assistant_message(response)

    question = input()
    prompt.add_user_assistant_message(KeenChatterInput(question=question))
```

It's a bit counter intuitive. Given current prompt api, I'd expect sth like:

```python
prompt = KeenChatterPrompt(KeenChatterInput(tone="nice", ...))  # render just the system message + user message optionally depending on input

while True: 
    question = input()
    prompt = prompt(KeenChatterInput(question=question))  # put user message rendering in __call__
    response = await llm.generate(prompt)

    prompt.add_user_assistant_message(KeenChatterInput(question=question))
    prompt.add_assistant_assistant_message(response)
```

This would work even in a current base case where we do one pass with:


```python
question = input()
prompt = KeenChatterPrompt(KeenChatterInput(tone="nice", question=question)) 
response = await llm.generate(prompt)
```

We can go even further and let the llm fill the history for us which would simplify the code even further:

```python
prompt = KeenChatterPrompt(KeenChatterInput(tone="nice", ...))  # render just the system message + user message optionally depending on input

while True: 
    question = input()
    prompt = prompt(KeenChatterInput(question=question))  # put user message rendering in __call__
    response = await llm.generate(prompt)  # llm.generate updates prompt chat before response return
```

mhordynski: Let's discuss any api changes to the prompt itself in different issue, right now this adds history support in compatible way without changing when we render user prompt.

PR #484, opened 2025-04-01
chore: remove git directory after tests

mhordynski:


pawel-chmielak-deepsense (approved):

packages/ragbits-document-search/tests/integration/test_git_source.py
```diff
@@ -14,6 +15,18 @@
 TEST_REPO_SSH_URL = "git@github.com:psf/requests.git"
 
 
+@pytest.fixture(autouse=True)
+def cleanup_test_dirs():
```
ludwiktrammer, line 19: Why are we forcing the local storage dir to be inside our repository in the first place? Without the `os.environ[LOCAL_STORAGE_DIR_ENV] = Path(__file__).parent.as_posix()` line it would be created in a temporary directory, which seems like a saner solution.

PR #487, opened 2025-04-02
chore: make use of temporary directory for all source tests

mhordynski:


packages/ragbits-document-search/tests/conftest.py
```diff
@@ -0,0 +1,16 @@
...
+
+from ragbits.document_search.documents.sources.base import LOCAL_STORAGE_DIR_ENV
+
+
+@pytest.fixture(scope="module", autouse=True)
+def configure_local_storage_dir():
+    random_tmp_dir = Path(tempfile.mkdtemp())
+    os.environ[LOCAL_STORAGE_DIR_ENV] = random_tmp_dir.as_posix()
+    yield
+    shutil.rmtree(random_tmp_dir)
```
ludwiktrammer, line 16: Non-blocking:

This should behave the same as the longer version:
```suggestion
def configure_local_storage_dir(tmpdir):
    os.environ[LOCAL_STORAGE_DIR_ENV] = tmpdir
```

micpst: there is a scope mismatch  for `tmpdir` fixture

micpst (approved):

PR #489, opened 2025-04-02
feat: Chat history context init

jakubduda-dsai:


ui/src/contexts/HistoryContext/HistoryContext.tsx
```diff
@@ -0,0 +1,37 @@
...
+);
+
+export const ChatHistoryProvider: React.FC<{ children: ReactNode }> = ({
+  children,
+}) => {
+  const [messages, setMessages] = useState<ChatMessageProps[]>([]);
+
+  const addMessage = (message: ChatMessageProps) => {
+    setMessages((prevMessages) => [...prevMessages, message]);
+  };
```
dazy-ds, line 16: I wonder whether this context should also handle creation of messages. We might for example expose methods such as: `createMessage` and `updateMessage` which could be helpful in case of streaming of adding a reference to existing message. `createMessage` could for example build a message, giving it an unique id, which would be returned. Then we would store some lookup for the messages along with the classic array, to allow easy updates.

That would ensure that the message creation is handled in one place and could be beneficial for users of this code who otherwise might need to write own implementations in every new project.

WDYT?

dazy-ds: Yeah, let's start with that. We can always refine in the future.

ui/src/contexts/HistoryContext/HistoryContext.tsx
```diff
@@ -0,0 +1,51 @@
...
+      const index = updatedMessages.findIndex((msg) => msg.id === id);
+      if (index !== -1) {
+        updatedMessages[index] = {
+          ...updatedMessages[index],
+          message: updatedMessages[index].message + data,
+        };
+      }
+
+      return updatedMessages;
+    });
```
dazy-ds, line 30: I think that the id should be generated inside of `createMessage` by the context, not by the user (we can later provide a way to modify id generation strategy).

The other thing is that updating messages this way could be really expensive (imagine chat implementation streaming token by token in a very long chat). Let's add some lookup for easier access (avoiding `.findIndex`)

ui/src/App.tsx
```diff
@@ -95,9 +81,12 @@ export default function Component() {
   return (
     <div className="h-full w-full max-w-full">
       <Layout subTitle="by deepsense.ai" title="Ragbits Chat">
+        <Button color="primary" onPress={clearMessages}>
+          Clear chat
+        </Button>
         <div className="relative flex h-full flex-col overflow-y-auto p-6 pb-8">
           <ScrollShadow className="flex h-full flex-col gap-6">
-            {messages.map((message, idx) => (
+            {Array.from(messages.values()).map((message, idx) => (
```
dazy-ds, line 89: nit: We can expose `messages` as array for the consumers to be more convenient

dazy-ds (approved):

PR #495, opened 2025-04-03
feat: integration tests for pgvector

kzamlynska:


kzamlynska:
pgvector integration tests can't be run with other vector stores.
I need to add image embeddings and refactor code a bit so it will be possible.
After refactoring test for pgvector will be merged with others.

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -362,7 +334,8 @@ async def retrieve(
                     VectorStoreResult(
                         entry=VectorStoreEntry(
                             id=record["id"],
-                            text=record["key"],
+                            text=record["text"] or "",
```
ludwiktrammer, line 337: Why is ` or ""` needed here?

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
```diff
@@ -393,5 +366,20 @@ async def list(
...
         with trace(table=self._table_name, query=where, limit=limit, offset=offset) as outputs:
             list_query, values = self._create_list_query(where, limit, offset)
-            outputs.listed_entries = await self._fetch_records(list_query, values)
+            try:
+                async with self._client.acquire() as conn:
+                    results = await conn.fetch(list_query, *values)
+                outputs.listed_entries = [
+                    VectorStoreEntry(
+                        id=record["id"],
+                        text=record["text"] or None,
```
ludwiktrammer, line 375: Why the `or None`?

kzamlynska: removed.

packages/ragbits-core/src/ragbits/core/vector_stores/pgvector.py
kzamlynska: You are right, we don't need it.  I removed it.

PR #506, opened 2025-04-08
feat: separate forms like/dislike + api config route

jakubduda-dsai:


ui/src/App.tsx
```diff
@@ -51,6 +61,34 @@ export default function Component() {
...
+
+  useEffect(() => {
+    if (config) {
+      if (
+        config.hasOwnProperty("like_form") ||
+        config.hasOwnProperty("dislike_form")
+      ) {
+        pluginManager.activate(FeedbackFormPluginName);
+      }
+    }
```
dazy-ds, line 89: nit: I am not a fan of that much nesting. We can change it to use be something like this:
```
if (!config || (!config.hasOwnProperty("likeForm) && !config.hasOwnProperty("dislike_form")) {
 return
}
```
We can even extract the second part to named variable, like `hasFeedbackForms` to give it more meaning

packages/ragbits-api/src/ragbits/api/_main.py
```diff
@@ -17,6 +19,8 @@
 
 logger = logging.getLogger(__name__)
 
+STARTED_FROM_DIR = Path(os.getcwd()).resolve()
```
dazy-ds, line 22: nit: Maybe let's change the name of this variable to `WORKING_DIRECTORY`. We can also consider removing it, as it's only used in one place and can be inlined in line 57 (that way maybe we would only call `resolve` once, but I am not sure if it's needed at all)

ui/src/App.tsx
```diff
@@ -9,26 +9,36 @@ import {
...
 import remarkGfm from "remark-gfm";
 import { Icon } from "@iconify/react/dist/iconify.js";
+import axiosWrapper from "./core/utils/axiosWrapper.ts";
+import { FormSchema } from "./plugins/FeedbackFormPlugin/types.ts";
 
 export default function Component() {
   const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [config, setConfig] = useState<{
+    like_form?: FormSchema | null;
+    dislike_form?: FormSchema | null;
```
dazy-ds, line 26: Maybe let's use an enum for the names, something like:
```
enum FormType {
 LIKE "like_form"
 DISLIKE: "dislike_form"
}
```
The goal is to use it to define type describing this state and chosen form in the lines 31-33 and hide the snake cased string literals.

ui/src/App.tsx
```diff
@@ -9,26 +9,36 @@ import {
...
+  } | null>(null);
   const [message, setMessage] = useState<string>("");
-  const scrollContainerRef = useRef<HTMLDivElement | null>(null);
-  const cancelRef = useRef<(() => void) | null>(null);
   const [showScrollDownButton, setShowScrollDownButton] = useState(false);
   const [shouldAutoScroll, setShouldAutoScroll] = useState(true);
+  const [feedbackName, setFeedbackName] = useState<
+    "like_form" | "dislike_form"
+  >();
```
dazy-ds, line 33: ^ Same as above, let's hide the snake cased string literals

ui/src/App.tsx
```diff
@@ -51,6 +61,34 @@ export default function Component() {
     }
   }, []);
 
+  useEffect(() => {
+    const handleFetchConfig = async () => {
+      setIsLoading(true);
+
+      const [data] = await axiosWrapper<Record<string, unknown>>({
+        url: "http://127.0.0.1:8000/api/config",
```
dazy-ds, line 69: Let's use `buildApiUrl` method with absolute path -> `buildApiUrl('/api/config')` (might require rebasing)

ui/src/App.tsx
```diff
@@ -51,6 +61,34 @@ export default function Component() {
     }
   }, []);
 
+  useEffect(() => {
+    const handleFetchConfig = async () => {
+      setIsLoading(true);
+
+      const [data] = await axiosWrapper<Record<string, unknown>>({
```
dazy-ds, line 68: We know the type of the config returned from the server, let's type it in the `types/api.ts` and use the type here.

ui/src/core/components/ChatMessage.tsx
```diff
@@ -8,15 +8,27 @@ import { Icon } from "@iconify/react";
 import DelayedTooltip from "./DelayedTooltip";
 import { useThemeContext } from "../../contexts/ThemeContext/useThemeContext.ts";
 import { Theme } from "../../contexts/ThemeContext/ThemeContext.ts";
+import { FormSchema } from "../../plugins/FeedbackFormPlugin/types.ts";
 
 export type ChatMessageProps = {
   classNames?: string[];
   chatMessage: ChatMessage;
-  onOpenFeedbackForm?: () => void;
+  onOpenFeedbackForm?: (name: "like_form" | "dislike_form") => void;
```
dazy-ds, line 16: Mentioned above, let's also use `enum` here

dazy-ds (commented):
Overall looks good, some minor changes from my side.

One thing that should be added to the backlog is creating a store/context with the config so that is accessible everywhere and fetched before app loads.

Regarding the enums, you can use them or create union type e.g. `type FormType = ...` and object that reassembles the enum (if you don't want to use them)

dazy-ds (approved):
Let's fix linters in Python code, and we can merge

PR #508, opened 2025-04-11
feat: LocalEmbedder to use sentence-transformers instead of torch

amaciaszek-dsai:


packages/ragbits-core/pyproject.toml
```diff
@@ -51,6 +51,7 @@ chroma = [
     "chromadb>=0.6.3,<1.0.0",
 ]
 local = [
+    "sentence-transformers==4.0.2",
```
mhordynski, line 54: Pinning versions like that makes it hard to install ragbits in projects with big number of dependencies. 
```suggestion
    "sentence-transformers>=4.0.2,<5.0.0",
```

mhordynski (approved):

packages/ragbits-core/src/ragbits/core/embeddings/local.py
```diff
@@ -74,26 +76,12 @@ async def embed_text(self, data: list[str], options: LocalEmbedderOptions | None
             data=data,
             model_name=self.model_name,
             model_obj=repr(self.model),
-            tokenizer=repr(self.tokenizer),
-            device=self.device,
             options=merged_options.dict(),
         ) as outputs:
             embeddings = []
             for batch in self._batch(data, merged_options.batch_size):
```
micpst, line 82: Do we still need `batch_size` here? afaik, you can  directlynprovide `batch_size` to `self.model.encode`

akotyla (approved):

packages/ragbits-core/src/ragbits/core/embeddings/local.py
amaciaszek-dsai: Right, I removed batch iteration

packages/ragbits-core/src/ragbits/core/embeddings/local.py
```diff
@@ -20,6 +20,8 @@ class LocalEmbedderOptions(Options):
     """
 
     batch_size: int = 1
+    model_init_kwargs: dict[str, Any] = field(default_factory=dict)
```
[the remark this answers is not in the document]
amaciaszek-dsai: I modified it to allow `**kwargs` when initiating the class

micpst (approved):

PR #509, opened 2025-04-14
feat: add docling document parser

micpst:


micpst:
@kzamlynska is it slower than unstructured?

kzamlynska:
> @kzamlynska is it slower than unstructured?

I don't know. I have problems installing unstructured because of dependencies conflicts, and I don't have API key for it. It is much complicated to run unstructured so docling is a winner here.

Here are times I've got for tests:

PASSED            [ 33%]Duration: 0.02 seconds
PASSED               [ 66%]Duration: 82.80 seconds
PASSED                 [100%]Duration: 14.65 seconds

micpst:
interesting, in my case it's ~13s for pdfs and images

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/docling.py
```diff
@@ -0,0 +1,141 @@
...
+        DocumentType.PPTX,
+        DocumentType.XLSX,
+        DocumentType.MD,
+        DocumentType.PNG,
+        DocumentType.JPG,
+        DocumentType.HTML,
+        DocumentType.PDF,
+    }
+
+    def __init__(self, ignore_images: bool = False, num_threads: int = 1) -> None:
```
mhordynski, line 36: not blocking, but I think we should have one proper way of saying what element I want to ingest into document search instance. Right now I think there are at least two ways of doing so (here, noop enricher).

mhordynski (approved):

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/docling.py
micpst: parsers are responsible for that, enrichers work on elements, but don't impact the final element types to be ingested. if you want to filter out some element types, parser is a right place to do so.

PR #515, opened 2025-04-15
chore: cursor rules

mhordynski:


.cursor/rules/repository-setup.mdc
```diff
@@ -0,0 +1,16 @@
+---
+description: 
+globs: 
+alwaysApply: true
+---
+
+# Repository setup and pre-commit checks
+
+This repository is using `uv` for package management. Rather than using `pip` use `uv pip` to install any packages or run python.
```
ludwiktrammer, line 9:
```suggestion
This repository is using `uv` for package management. Rather than using `pip` use `uv add` to install any packages and `uv run` to run python.
```

.pre-commit-config.yaml
```diff
@@ -7,7 +7,7 @@ repos:
           - id: check-case-conflict
           - id: check-merge-conflict
           - id: trailing-whitespace
-            exclude: .bumpversion.cfg|notebooks/.*\.py
+            exclude: .bumpversion.cfg|notebooks/.*\.py|.cursor/
```
micpst, line 10: this exclude can be removed altogether

mhordynski: nah, cursor has its own "edit" view for .mdc files where header (`description`, `globs` and `alwaysApply`) is populated from rendered form. It adds trailing whitespaces for some reason and it's annoying without this exclude.

We can ditch bumpversion and noteboooks though

.cursor/rules/repository-setup.mdc
mhordynski: Cursor often installs temporary packages to test something. That's why I've decided to put `uv pip` rather than `uv add` here. 

Thanks for `uv run` suggestion - applied.

.pre-commit-config.yaml
micpst: ok, got it

micpst (approved):

PR #518, opened 2025-04-17
feat: add e2e chatbot API and UserInterface

mhordynski:


examples/api/chat.py
micpst, line 1: This example does nothing. Let's add maybe a docstring similar to what we have in other examples so the user'll know how to run this and add `api.run()` to setup the whole thing.
```python
from ragbits.api._main import RagbitsAPI

api = RagbitsAPI(
    chat_interface=MyChat,
    cors_origins=["http://localhost:3000"],  # Optional: Add CORS origins if needed
    ui_build_dir=None,  # Optional: Path to custom UI build
)
api.run(host="127.0.0.1", port=8000)
```

packages/ragbits-api/src/ragbits/api/__init__.py
micpst, line 1: We should export all public classes from `_main.py` here

examples/api/chat.py
mhordynski: Fixed

packages/ragbits-api/src/ragbits/api/__init__.py
mhordynski: After merging ragbits-conversations and ragbits-api to ragbits-chat - I've moved `_main` contents into `ragbits.chat.api` package

micpst (approved):

PR #528, opened 2025-04-23
feat: Allow using sparse embeddings with Qdrant and local vector stores

ludwiktrammer:


micpst (commented):
PR looks good to me. I'd just split the dense and sparse embeddings into separate modules inside `embeddings` module for better readability.

ludwiktrammer:
@micpst Makes sense! Will do.

Another think that would make a lot of sense would be changing the names of the classes so that the inheritance hierarchy looks like that:

```
* Embedder
    * DenseEmbedder
    * SparseEmbedder
```

I haven't done this in this PR because I didn't want to suddenly create a lot of backwards incompatible chaos. But I think it's worth thinking about long term.

ludwiktrammer:
@micpst Done

micpst:
@ludwiktrammer yeah, I agree, we can do it, even in this PR. As for backward compatibility, we still don't have a stable release, so there will be no better time than now to do it. If you still have time you can do it in this PR

ludwiktrammer:
@micpst Okay, done :)

micpst (approved):

packages/ragbits-core/tests/integration/vector_stores/test_vector_store_sparse.py
```diff
@@ -0,0 +1,123 @@
...
+        else [e for e in vector_store_entries if e.image_bytes is not None]
+    )
+
+    sorted_results = sorted(result_entries, key=lambda r: r.entry.id)
+    sorted_expected = sorted(expected_entries, key=lambda entry: entry.id)
+
+    prev_score = float("inf")
+    for result, expected in zip(sorted_results, sorted_expected, strict=True):
+        assert result.entry.id == expected.id
+        print(result.score, result.entry.text, result.vector)
```
mhordynski, line 99: Debug print left behind

mhordynski (approved):

PR #529, opened 2025-04-24
feat: integrate sources with dataloaders

micpst:
This PR is part of `evaluation` package refactor.

Changes:
- improves typing in evaluation package
- integrates sources with data loaders
- changes hugging face source allowing it to download whole table instead of single row
- allows sources to be constructed from configs (this is required by data loaders)


packages/ragbits-evaluate/src/ragbits/evaluate/dataloaders/document_search.py
```diff
@@ -0,0 +1,33 @@
...
+            The document search evaluation data.
+        """
+        data_path = await self.source.fetch()
+        dataset = load_dataset(
+            path=str(data_path.parent),
+            split=data_path.stem,
+        )
+        return [
+            DocumentSearchData(
+                question=data["question"],
```
mhordynski, line 29: Let's explicitly check for "question" and "chunks" before that line and raise with error explaining what structure we should have in the dataset, maybe send to how-to guide in our documentation.

packages/ragbits-evaluate/src/ragbits/evaluate/dataloaders/document_search.py
```diff
@@ -0,0 +1,33 @@
...
+                reference_passages=data["chunks"],
```
mhordynski, line 30: Shouldn't be "passages"? If not then we have a mistake in docstring

packages/ragbits-core/src/ragbits/core/sources/hf.py
```diff
@@ -32,9 +37,8 @@ def id(self) -> str:
         Returns:
             Unique identifier.
         """
-        return f"{self.protocol}:{self.path}/{self.split}/{self.row}"
+        return f"{self.protocol}:{self.path}/{self.split}" + f"/{self.row}" if self.row is not None else ""
```
mhordynski, line 40: nit: I don't like having string concat with ternary operator mixed, easy to make errors

```suggestion
return f"{self.protocol}:{self.path}/{self.split}{f'/{self.row}' if self.row is not None else ''}"
```

packages/ragbits-core/src/ragbits/core/sources/hf.py
```diff
@@ -17,11 +17,16 @@
 class HuggingFaceSource(Source):
     """
     An object representing a Hugging Face dataset source.
+
+    Supports two formats:
+    1. Complete dataset: When no row is specified, downloads the entire dataset. Used for QA datasets.
+    2. Single row: When a specific row is specified, downloads only that row. Used for document datasets
```
mhordynski, line 23: maybe it's different source then? Right now we have two very different behaviours in one class

packages/ragbits-evaluate/src/ragbits/evaluate/dataloaders/base.py
```diff
@@ -1,18 +1,49 @@
...
+        """
+        Create an instance of `DataLoader` from a configuration dictionary.
+
+        Args:
+            config: A dictionary containing configuration settings for the data loader.
+
+        Returns:
+            An instance of the data loader class initialized with the provided configuration.
+        """
+        source_config = ObjectConstructionConfig.model_validate(config["source"])
```
mhordynski, line 41: I think we should start to raise proper errors with explicitly saying what is happening. Right now this will fail with KeyError when "source" is not config

packages/ragbits-core/src/ragbits/core/sources/hf.py
micpst: yeah, i agree. The second case is super rare, and we support it only for demo purposes, cuz we host one dataset in such a format, normally we don't need this. I will split them for now

micpst: After thinking about it, maybe stay with one source, having two sources for the same platform can be confusing for the user

mhordynski (approved):

PR #550, opened 2025-05-09
chore: setup ci/cd for user interface

jakubduda-dsai:


mhordynski (approved):
lgtm

.github/workflows/prepare_release.yml
```diff
@@ -46,6 +51,14 @@ jobs:
           echo new_version=`grep version packages/${{ github.event.inputs.packageName }}/pyproject.toml | cut -d \" -f2` >> $GITHUB_OUTPUT
           uv sync
 
+      - name: Install UI dependencies
+        run: npm i
+        working-directory: ui
+
+      - name: Build UI
```
micpst, line 58: I know we discussed it yesterday, but I think however we should put this "build with commit" during merge to main. Imagine if someone wants to use ragbits from the main branch instead of pypi, they will have an outdated frontend. I'd make this additional commit only on merge/commit to main (if ui changed) and preserve the original author of the commit

cc @mhordynski

micpst (approved):

PR #552, opened 2025-05-12
feat: FE API integration improvements

dazy-ds:


jakubduda-dsai (approved):
I updated codebase from main because CI has changed + pushed ui build to repo so CI checks will pass

Changes looks good

PR #555, opened 2025-05-13
refactor: adjust eval piplines interface to batch mode

micpst:


packages/ragbits-evaluate/src/ragbits/evaluate/evaluator.py
```diff
@@ -114,17 +115,12 @@ async def _call_pipeline(
...
-        semaphore = asyncio.Semaphore(self.batch_size)
-
-        async def _call_pipeline_with_semaphore(data: EvaluationDataT) -> EvaluationResultT:
-            async with semaphore:
-                return await pipeline(data)
-
         start_time = time.perf_counter()
-        pipe_outputs = await tqdm.gather(*[_call_pipeline_with_semaphore(data) for data in dataset], desc="Evaluation")
+        pipe_outputs = await tqdm.gather(
+            *[pipeline(data) for data in batched(dataset, self.batch_size)], desc="Evaluation"
```
mhordynski, line 120: shouldn't we somehow handle and retry exceptions here? In case that one item of batch fails we lose all batch progress now

micpst: added error handling + errors dump for debugging

mhordynski (approved):

PR #557, opened 2025-05-14
feat: clear HeroUI components

jakubduda-dsai:
prompt in minimalistic version, removed unnecessary components based on HeroUI Pro


dazy-ds (changes requested):
In my opinion it's not enough.

Let's arrange some meeting to discuss this, when you have time.

dazy-ds (approved):

PR #558, opened 2025-05-14
feat: Update TailwindCSS

jakubduda-dsai:
Do not merge for now, this is beta of heroui but solution works, lets wait for official release of heroui version which supports TailwindCSS v4


dazy-ds (approved):
LGTM, let's just fix Prettier errors

PR #582, opened 2025-05-27
feat: init TypeScript Ragbits API Client modules

jakubduda-dsai:


typescript/ragbits-api-client-react/src/hooks.ts
```diff
@@ -0,0 +1,127 @@
...
+
+/**
+ * Generic hook for making API calls to Ragbits endpoints
+ * @param endpoint - The API endpoint (e.g., "/api/config", "/api/feedback")
+ * @param defaultOptions - Default options for the API call
+ */
+export function useRagbitsCall<T = any>(
+  endpoint: string,
+  defaultOptions?: ApiRequestOptions
+): RagbitsCallResult<T> {
```
dazy-ds, line 18: Since we know the predefined routes and their default return values, we can improve this type to automatically assign response types if endpoint is e.g. `/api/config`.

This would simplify the usage for predefined routes from ` useRagbitsCall<Config>("/api/config");` to `useRagbitsCall("/api/config");`

typescript/ragbits-api-client-react/src/hooks.ts
```diff
@@ -0,0 +1,127 @@
...
+    setIsLoading(false);
+  }, []);
+
+  return {
+    data,
+    error,
+    isLoading,
+    call,
+    reset,
+  };
```
dazy-ds, line 61: We can maybe think about adding a support for aborting the request using `AbortController`. It can be useful for user's that invoke the `call` function inside `useEffects`.

typescript/ragbits-api-client-react/src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+export interface RagbitsStreamResult<T> {
+  isStreaming: boolean;
+  error: Error | null;
+  stream: (
+    endpoint: string,
+    data: any,
+    callbacks: StreamCallbacks<T>
+  ) => () => void;
+  cancel: () => void;
+}
```
dazy-ds, line 41: Maybe we can also expose `Error` type as generic so that the users can define their error types? So we would for example have `RagbitsCallResult<T, E = Error>...`

It might not be needed now, but I think it's a helpful addition.

typescript/ragbits-api-client/src/index.ts
```diff
@@ -0,0 +1,198 @@
...
+        await processStream(response);
+      } catch (error) {
+        console.error("Request error:", error);
+        const errorMessage =
+          error instanceof Error ? error.message : "Error connecting to server";
+        await callbacks.onError(errorMessage);
+      }
+    };
+
+    void startStream();
```
dazy-ds, line 189: We should catch any errors here and call `onError` in case they happen.

typescript/ui/src/App.tsx
```diff
@@ -40,7 +38,17 @@ export default function Component() {
     FeedbackFormPluginName,
   );
   const showHistory = useMemo(() => history.length > 0, [history.length]);
-  const configRequest = useRequest<ConfigResponse>(buildApiUrl("/api/config"));
+
+  // Use the new generic hooks
+  const config = useRagbitsCall<Record<string, unknown>>("/api/config");
```
dazy-ds, line 43: Shouldn't we use the `Config` type here?

typescript/ui/src/App.tsx
```diff
@@ -200,49 +201,47 @@ You can ask me anything! I can provide information, answer questions, and assist
...
-              tabIndex={-1}
-              startContent={<Icon icon="heroicons:arrow-down" />}
+              isIconOnly
+              variant="flat"
+              aria-label="Scroll to bottom"
+              className="absolute bottom-24 right-6 z-20"
+              onClick={scrollToBottom}
             >
-              Scroll to bottom
+              <Icon icon="heroicons:arrow-down-20-solid" />
```
dazy-ds, line 215: Can we use the same icon as before? If not, let's add this icon to be preloaded in the `main.tsx`

typescript/ui/src/contexts/HistoryContext/HistoryContextProvider.tsx
```diff
@@ -111,63 +108,50 @@ export function HistoryProvider({ children }: PropsWithChildren) {
...
+            {
+              type: ChatResponseType.TEXT,
+              content: error,
+            },
+            assistantResponseId,
+          );
+        },
+        onClose: () => {
+          // Stream closed
+        },
```
dazy-ds, line 149: Let's use the `noop` function from `lodash`

dazy-ds (changes requested):
I left some comments to be addressed.

I am also thinking about renaming the `typescript` directory to something more descriptive. Maybe `client` would be good? WDYT?

typescript/ragbits-api-client-react/src/hooks.ts
```diff
@@ -0,0 +1,210 @@
...
+  const [error, setError] = useState<Error | null>(null);
+  const [isLoading, setIsLoading] = useState(false);
+  const abortControllerRef = useRef<AbortController | null>(null);
+
+  const abort = useCallback(() => {
+    if (abortControllerRef.current) {
+      abortControllerRef.current.abort();
+      abortControllerRef.current = null;
+      setIsLoading(false);
+    }
```
dazy-ds: nit: I am more of a fan of early returns, rather than nesting. In that case let's just decide one approach and stick to it across the project.

I suggest using `if (!abortControllerRef.current) { return } ...rest of the code...`, but keeping it as is also seems fine to me.

typescript/ragbits-api-client-react/src/hooks.ts
```diff
@@ -0,0 +1,210 @@
...
+  }, []);
+
+  const stream = useCallback(
+    (
+      data: StreamingEndpointRequest<TEndpoint>,
+      callbacks: StreamCallbacks<TResponse, string>
+    ): (() => void) => {
+      // Cancel any existing stream
+      cancel();
```
dazy-ds: nit: In `useRagbitsCall` you are calling AbortController's `abort` directly, without causing rerender caused by `setIsLoading` call. Let's decide on one convention. Either we skip the rerender and call the controller directly here, or use `abort` function in the `useRagbitsCall` above.

I think that we shouldn't cause excessive rerenders when it's not needed. Here we are setting `isStreaming` to `false` to just change it to `true` after few instructions. Of course React can batch those calls, but in that case it's just dead code.

typescript/ragbits-api-client-react/src/hooks.ts
```diff
@@ -0,0 +1,210 @@
...
+
+      const cancelFn = (
+        client as {
+          makeStreamRequest<T>(
+            endpoint: string,
+            data: StreamingEndpointRequest<TEndpoint>,
+            callbacks: StreamCallbacks<T, Error>,
+            signal?: AbortSignal
+          ): () => void;
+        }
```
dazy-ds: I think this type cast is not needed. `client` variable should be typed from the definition.
Please correct me if I'm wrong, otherwise let's remove it

typescript/ragbits-api-client/src/index.ts
```diff
@@ -0,0 +1,202 @@
...
+
+        console.error("Request error:", error);
+        const errorMessage =
+          error instanceof Error ? error.message : "Error connecting to server";
+        await callbacks.onError(new Error(errorMessage));
+      }
+    };
+
+    try {
+      void startStream();
```
dazy-ds: I'm not sure about usage of `void` here. I would remove it, if the linter can't detect the `try/catch` block, let's change it to `.catch`.

typescript/ui/src/contexts/HistoryContext/HistoryContextProvider.tsx
```diff
@@ -1,28 +1,23 @@
...
+} from "ragbits-api-client-react";
+import { ChatMessage, HistoryState } from "../../types/history.ts";
+import { mapHistoryToMessages } from "../../core/utils/messageMapper";
 
 export function HistoryProvider({ children }: PropsWithChildren) {
   const [history, setHistory] = useState<HistoryState>(new Map());
   const [serverState, setServerState] = useState<ServerState | null>(null);
   const [conversationId, setConversationId] = useState<string | null>(null);
-  const [unsubscribe, setUnsubscribe] = useState<UnsubscribeFn>(null);
+  const stream = useRagbitsStream("/api/chat");
```
dazy-ds: nit: I would change the variable name to something like `streamFactory`. Or maybe we should rename the `.stream` method to  `.start` or `.initialize`. I am just not convinced about the `stream.stream(...)` notation.

typescript/ui/src/core/utils/messageMapper.ts
```diff
@@ -0,0 +1,12 @@
+import { Message } from "ragbits-api-client-react";
+import { ChatMessage } from "../../types/history";
+
+/**
+ * Maps the internal history messages to the API message format
+ */
+export function mapHistoryToMessages(messages: ChatMessage[]): Message[] {
```
dazy-ds: nit: We can change the argument type to `HistoryState` and extract the underlying values here. That would allow us to simplify the call to `mapHistoryToMessages(history)` above.

dazy-ds (approved):
Just some minor nits. Looks good overall.

jakubduda-dsai (commented):
Rebuild UI before merging to be sure there is the newest codebase build & merged

PR #591, opened 2025-06-02
feat: support pdf files in prompt

mhordynski:


packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -286,29 +283,38 @@ async def _get_litellm_response(
...
+            # Router instances have these configured at initialization time
+            if self.router is None:
+                if self.api_base is not None:
+                    completion_kwargs["base_url"] = self.api_base
+                if self.api_key is not None:
+                    completion_kwargs["api_key"] = self.api_key
+                if self.api_version is not None:
+                    completion_kwargs["api_version"] = self.api_version
+
+            response = await entrypoint.acompletion(**completion_kwargs)
```
ds-juliamay, line 305: This edit was caused by change in litellm acompletion signature?

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -317,6 +323,37 @@ def _get_response_format(
...
+            has_images = False
+            has_pdfs = False
+
+            for message in prompt.chat:
+                content = message.get("content", [])
+                if isinstance(content, list):
+                    for item in content:
+                        if item.get("type") == "image_url":
+                            has_images = True
+                        elif item.get("type") == "file":
```
ds-juliamay, line 349: Does type "file" always mean that it is a pdf file or it can also be some other type of file?

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -273,36 +282,109 @@ def add_assistant_message(self, message: str | PromptOutputT) -> "Prompt[PromptI
...
+        """
+        Returns the attachments in form of URLs or base64 encoded strings.
+
+        Returns:
+            list[str]: List of attachment URLs/base64 strings for both images and PDFs
+        """
+        attachments = []
+        for message in self.chat:
+            content = message["content"]
+            if not isinstance(content, list):
```
ds-juliamay, line 295: If we have only one attachment then it a list with one item?

packages/ragbits-core/tests/unit/prompts/test_prompt.py
```diff
@@ -4,7 +4,7 @@
 import pytest
```
ds-juliamay, line 4: Do we also need unit tests for prompts with pdf files?

packages/ragbits-core/src/ragbits/core/llms/litellm.py
mhordynski: It was related to change in Router.acompletion signature - due to bump of litellm lib

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
mhordynski: yes, this is always a list with objects

PR #600, opened 2025-06-04
feat: add function calling to ragbits llm

ds-juliamay:


examples/core/llms/tool_use.py
```diff
@@ -0,0 +1,93 @@
...
+    Returns the current weather for a given location.
+
+    Args:
+        location: The location to get the weather for.
+
+    Returns:
+        The current weather for the given location.
+    """
+    if "tokyo" in location.lower():
+        return json.dumps({"location": "Tokyo", "temperature": "10", "unit": "celsius"})
```
ghost: signature of the functions says that we should return dict while this returns str

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -147,25 +150,54 @@ async def _call(
...
 
         response_format = self._get_response_format(output_schema=output_schema, json_mode=json_mode)
 
-        start_time = time.perf_counter()
-        response = await self._get_litellm_response(
-            conversation=prompt.chat,
-            options=options,
-            response_format=response_format,
-        )
+        if tools:
```
ghost: why is this if needed? it seems that `self._get_litellm_response` accepts tools as well with None being a valid option

packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -222,25 +256,40 @@ async def _call_streaming(
...
-            stream=True,
-        )
-        if not response.completion_stream:  # type: ignore
+        if tools:
+            start_time = time.perf_counter()
+            response = await self._get_litellm_response(
+                conversation=prompt.chat,
+                options=options,
+                response_format=response_format,
+                tools=tools,
```
ghost: same as above, it seems that _get_litellm_response accepts tools

docs/how-to/llms/use_tools_with_llms.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Using tools with LLMs
+
+Tools for LLMs need to be defined as functions with docstrings describing the function's purpose, its arguments and its return value:
+
+```python
+from ragbits.core.prompt import Prompt
+
+
+def get_weather(location: str, units: str = None) -> dict:
```
ghost: shouldn't this return str?

examples/core/llms/tool_use.py
ds-juliamay: Yes, it should be annotated as string, thanks for pointing this out

docs/how-to/llms/use_tools_with_llms.md
ds-juliamay: It should return string, I'll fix this type annotation

packages/ragbits-core/src/ragbits/core/llms/litellm.py
ds-juliamay: I think it is not necessary here, I'll remove it

packages/ragbits-core/src/ragbits/core/llms/litellm.py
ds-juliamay: Same as above

mkdocs.yml
```diff
@@ -17,6 +17,7 @@ nav:
       - LLMs:
         - "Interact with LLMs": how-to/llms/use_llms.md
         - "Use local or self-hosted LLMs": how-to/llms/use_local_llms.md
+        - "Use tools with LLMs": how-to/llms/use_tools_with_llms.md
```
micpst, line 19: nit: lets move it above "Use local or self-hosted LLMs" how-to, its more common to use tools than setting up local LLMs

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -127,38 +167,79 @@ async def generate(
         prompt: BasePromptWithParser[PromptOutputT],
         *,
         options: LLMClientOptionsT | None = None,
+        tools: None = None,
     ) -> PromptOutputT: ...
 
+    @overload
```
micpst: this overload is not needed

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -127,38 +167,79 @@ async def generate(
...
     @overload
     async def generate(
         self,
         prompt: BasePrompt,
         *,
         options: LLMClientOptionsT | None = None,
+        tools: None = None,
     ) -> PromptOutputT: ...
 
+    @overload
```
micpst, line 157: ditto

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -127,38 +167,79 @@ async def generate(
...
+        tools: Tools | None = None,
+    ) -> PromptOutputT | ToolCallsResponse: ...
+
     @overload
     async def generate(
         self,
         prompt: BasePrompt,
         *,
         options: LLMClientOptionsT | None = None,
+        tools: None = None,
```
micpst: just type it as `Tools | None = None`

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -168,27 +249,54 @@ async def generate(
...
             if isinstance(prompt, BasePromptWithParser):
                 response = await prompt.parse_response(raw_response["response"])
             else:
                 response = cast(PromptOutputT, raw_response["response"])
             raw_response["response"] = response
             outputs.response = raw_response
+            return response
 
-        return response
+    @overload
```
micpst: no need to overload

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -197,6 +305,25 @@ async def generate_with_metadata(
         prompt: BasePrompt,
         *,
         options: LLMClientOptionsT | None = None,
+        tools: None = None,
+    ) -> LLMResponseWithMetadata[PromptOutputT]: ...
+
+    @overload
```
micpst: ditto

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -197,6 +305,25 @@ async def generate_with_metadata(
...
+    async def generate_with_metadata(
+        self,
+        prompt: BasePrompt,
+        *,
+        options: LLMClientOptionsT | None = None,
+        tools: Tools | None = None,
+    ) -> LLMResponseWithMetadata[PromptOutputT]: ...
+
+    @overload
```
micpst, line 208: ditto

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -214,13 +342,25 @@ async def generate_with_metadata(
         prompt: ChatFormat,
         *,
         options: LLMClientOptionsT | None = None,
+        tools: None = None,
+    ) -> LLMResponseWithMetadata[PromptOutputT]: ...
+
+    @overload
+    @overload
```
micpst: ditto

konrad-czarnota-ds (approved):
LGTM :+1: you just need to fix a conflict

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -30,13 +47,33 @@ class LLMType(enum.Enum):
     STRUCTURED_OUTPUT = "structured_output"
 
 
+class ToolCall(BaseModel):
+    """
+    A schema of tool call data
+    """
+
+    tool_name: str
+    tool_arguments: str
```
micpst: i think we can store parsed arguments here

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -30,13 +47,33 @@ class LLMType(enum.Enum):
...
+    tool_type: str
+    tool_call_id: str
```
micpst: these fields are provider specific. You can just add `model_config = ConfigDict(extra="allow")` to make it more generic

ds-juliamay: We should hide both tool_type and tool_call_id in this config dict and explicitly include only tool_name and tool_arguments?

packages/ragbits-core/src/ragbits/core/llms/base.py
ds-juliamay: Done, now they are stored as dict

packages/ragbits-core/src/ragbits/core/llms/base.py
ds-juliamay: I did it this way

mkdocs.yml
ds-juliamay: Done

micpst (approved):

PR #606, opened 2025-06-09
docs: Update use_pgVector_store.md

Pleto:
There should be no spaces around the equals sign. Docker interprets this as an invalid argument, which in turn causes the image name parsing to fail, resulting in the "invalid reference format" error


mhordynski (approved):

mhordynski:
Thank you for your contribution @Pleto :raised_hands: 
Merging into main, I'll deploy docs soon.

PR #610, opened 2025-06-10
feat: ragbits api clients unit tests

jakubduda-dsai:


typescript/ragbits-api-client-react/src/test/RagbitsProvider.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+        render(
+            <RagbitsProvider baseUrl="https://api.example.com">
+                <TestComponent />
+            </RagbitsProvider>
+        )
+
+        expect(screen.getByTestId('client-available')).toHaveTextContent(
+            'Client available'
+        )
+    })
```
dazy-ds: This test doesn't really check that the custom URL is really set. Let's either remove it or change the implementation to show the base link in the component (or other similar solution), so that we can really check if it's working.

typescript/ragbits-api-client-react/src/test/RagbitsProvider.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+        rerender(
+            <RagbitsProvider baseUrl="https://api2.example.com">
+                <TestComponent />
+            </RagbitsProvider>
+        )
+
+        expect(screen.getByTestId('client-available')).toHaveTextContent(
+            'Client available'
+        )
+    })
```
dazy-ds: Same as above, this test only validates that the prop can be set (which is checked by TypeScript ahead of time)

typescript/ragbits-api-client-react/src/test/index.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        expect(RagbitsReact.useRagbitsContext).toBeDefined()
+    })
+
+    it('should export RagbitsProvider', () => {
+        expect(RagbitsReact.RagbitsProvider).toBeDefined()
+    })
+
+    it('should export RagbitsClient', () => {
+        expect(RagbitsClient).toBeDefined()
+        expect(typeof RagbitsClient).toBe('function')
```
dazy-ds: I like the check for the type. Maybe let's it for tests above (all hooks, RagbitsProvider)?

typescript/ragbits-api-client-react/src/test/index.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+        expect(ChatResponseType.MESSAGE).toBe('message')
+
+        expect(FormFieldType).toBeDefined()
+        expect(FormFieldType.TEXT).toBe('text')
+        expect(FormFieldType.SELECT).toBe('select')
+
+        expect(FeedbackType).toBeDefined()
+        expect(FeedbackType.LIKE).toBe('like')
+        expect(FeedbackType.DISLIKE).toBe('dislike')
+    })
```
dazy-ds: We can keep the checks for `MessageRole` and `ChatResponseType` enums, as they must be synchronized with the Python API (although this test needs updating right now).

Could you also remove `FormFieldType` and `FeedbackType` from tests and also from `types.ts` as they are not used anymore (I forgot to remove them during form standardization, sorry)?

typescript/ragbits-api-client-react/src/test/index.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+})
+
+describe('Type Definitions', () => {
```
dazy-ds: I think all this whole `describe` essentially checks that the TypeScript's type system works correctly.

I think we should leave that to the TS, and just remove it. WDYT?

typescript/ragbits-api-client/src/test/mocks/handlers.ts
```diff
@@ -0,0 +1,118 @@
...
+                        ],
+                    },
+                },
+                dislike: {
+                    enabled: true,
+                    form: null,
+                },
+            },
+        })
+    }),
```
dazy-ds: As the form definition is updated, we should also change this mock, so that the tests pass.

typescript/ragbits-api-client-react/src/test/integration.test.tsx
```diff
@@ -0,0 +1,336 @@
...
+                                    type: 'select',
+                                    required: true,
+                                    options: ['Incorrect', 'Unclear', 'Other'],
+                                },
+                            ],
+                        },
+                    },
+                },
+            })
+        })
```
dazy-ds: The response format was changed during the form standardization. Let's extract this to some variable and also use it in the mock definition.

typescript/ragbits-api-client-react/src/test/integration.test.tsx
```diff
@@ -0,0 +1,336 @@
...
+    return function Wrapper({ children }: { children: React.ReactNode }) {
+        return (
+            <RagbitsProvider baseUrl="http://127.0.0.1:8000">
+                {children}
+            </RagbitsProvider>
+        )
+    }
+}
+
+describe('Integration Tests', () => {
```
dazy-ds: ### First please read this comment regarding the section. If we decide to keep these tests, then let's address other comments in this file.

The more I think about these tests, the more they resemble unit tests. In most cases, we're primarily verifying that the mocked response matches the expected (also mocked) result. This approach can be problematic, because if an external implementation changes and the tests aren’t updated accordingly, the mocks will continue to pass, giving us no indication that something is broken.

Maybe we should consider testing against a contract generated automatically after each change to the `ragbits-api` package. This would ensure that implementations are in sync.

The other solution is to perform the tests using real instance of a backend (e.g. in Docker Image).

typescript/ragbits-api-client-react/src/test/integration.test.tsx
```diff
@@ -0,0 +1,336 @@
...
+            expect(typeof result.current.data).toBe('object')
+            expect(directResponse).toEqual(result.current.data)
+            expect(directResponse).toHaveProperty('feedback')
+            expect(directResponse.feedback).toHaveProperty('like')
+            expect(directResponse.feedback).toHaveProperty('dislike')
+        })
+    })
+
+    describe('Performance and Memory Integration', () => {
+        it('should not cause memory leaks when using both packages', async () => {
```
dazy-ds: This test verifies the mock server's ability to handle multiple requests. However, I believe stress or load testing should be handled on the backend side if needed, as it's more relevant to actual service performance.

typescript/ragbits-api-client-react/src/test/integration.test.tsx
```diff
@@ -0,0 +1,336 @@
...
+                await act(async () => {
+                    await result.current.call()
+                })
+            }
+
+            expect(result.current.data).toHaveProperty('feedback')
+            expect(result.current.error).toBeNull()
+        })
+
+        it('should handle multiple concurrent operations efficiently', async () => {
```
dazy-ds: Same as above

typescript/ragbits-api-client-react/src/test/mocks/handlers.ts
```diff
@@ -0,0 +1,109 @@
...
+                                label: 'Reason',
+                                type: FormFieldType.SELECT,
+                                required: true,
+                                options: ['Incorrect', 'Unclear', 'Other'],
+                            },
+                        ],
+                    },
+                },
+            },
+        }
```
dazy-ds: As the form definition is updated, we should also change this mock, so that the tests pass.

(This is exactly the kind of situation I mentioned in the general comment about integration tests. Since we're using a mock service here, we're not actually verifying compatibility with the real backend. The only thing this test guarantees is that the frontend conforms to the mock, not that it works with the actual API.)

typescript/ragbits-api-client-react/src/test/useRagbitsCall.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+        expect(result.current.data).toBeNull()
+        expect(result.current.error).toBeNull()
+        expect(result.current.isLoading).toBe(false)
+        expect(typeof result.current.call).toBe('function')
+        expect(typeof result.current.reset).toBe('function')
+        expect(typeof result.current.abort).toBe('function')
+    })
+
+    it('should make successful API call', async () => {
```
dazy-ds, line 34: Going back to the comment about integration tests. This could be the valid use of the mocked service, as we just test that our hook calls some endpoint and parses the data.

Also the form definition should be updated.

typescript/ragbits-api-client/src/test/RagbitsClient.test.ts
```diff
@@ -0,0 +1,581 @@
...
+                                },
+                            ],
+                        },
+                    },
+                    dislike: {
+                        enabled: true,
+                        form: null,
+                    },
+                },
+            })
```
dazy-ds: Form schema needs to be updated.

typescript/ragbits-api-client/src/test/RagbitsClient.test.ts
```diff
@@ -0,0 +1,581 @@
...
+            await new Promise((resolve) => setTimeout(resolve, 100))
+
+            expect(errors).toHaveLength(1)
+            expect(errors[0].message).toBe('Synchronous fetch error')
+
+            // Restore original fetch
+            global.fetch = originalFetch
+        })
+
+        it('should handle aborted streams gracefully', async () => {
```
dazy-ds: This test seems to be almost identical to the `should handle stream with AbortSignal` one. Could we remove one of them?

typescript/ragbits-api-client/src/test/RagbitsClient.test.ts
```diff
@@ -0,0 +1,581 @@
...
+            )
+
+            // Wait to ensure the abort is processed
+            await new Promise((resolve) => setTimeout(resolve, 150))
+
+            // When aborted, no error should be called
+            expect(errors).toHaveLength(0)
+        })
+
+        it('should handle pre-aborted signal in error handling', async () => {
```
dazy-ds: Same as above. When we catch one `AbortSignal` we are catching every event of that kind. In my opinion there is no need to test different cases, as this is handled by the `signal` mechanism of the `fetch`.

dazy-ds (changes requested):
Some minor changes, but I also left a very important comment about the integration tests. I think we should talk through it, potentially with @mhordynski if needed.

typescript/ragbits-api-client-react/src/test/integration.test.tsx
jakubduda-dsai: For now i removed all integration.tests to simplify it for start let's discuss it after first npm beta release

PR #616, opened 2025-06-11
feat: standardize form creation

dazy-ds:
After talks with @jakubduda-dsai we have decided to standardize the form creation DSL. Instead of defining our own language, which could grow as the time goes, we've decided to go with the `JSONSchema` and library that support it.

This allows users to define feedback forms using simple `Pydantic` models without the need to learn our custom DSL. This is also useful on the frontend side, as we just need to handle specific input types when the need for them arises.

I also configuration fetching flow and contained all of the feedback form logic in its plugin.


typescript/ui/src/App.tsx
```diff
@@ -71,18 +47,12 @@ export default function App() {
     }
   }, []);
 
+  // TODO: Might consider moving all default plugins activation to ConfigContextProvider
```
jakubduda-dsai: I like that idea, what you think about doing it in this PR?

jakubduda-dsai: or maybe event some PluginContext which would have pure logic about enabling/disabling plugins?

typescript/ui/src/contexts/ConfigContext/ConfigContext.ts
```diff
@@ -0,0 +1,10 @@
+import { ConfigResponse } from "ragbits-api-client";
```
jakubduda-dsai: I think that we can import ConfigResponse from ragbits-api-client-react and possibly ragbits-api-client can be removed from dependencies in ui

typescript/ui/src/plugins/FeedbackPlugin/FeedbackForm.tsx
```diff
@@ -0,0 +1,160 @@
...
+    });
+  }, []);
+
+  if (!schema) {
+    return null;
+  }
+
+  return (
+    <>
+      {feedback.like !== null && (
```
jakubduda-dsai: I think that correct condition looking at ConfigResponse should be feedback.like.enabled, same for dislike. What do you think?

dazy-ds: Good catch. Changed!

typescript/ui/src/contexts/ConfigContext/ConfigContext.ts
dazy-ds: Yeah, that would be better. I believe that auto import is the culprit here :D  Changed.

typescript/ui/src/App.tsx
dazy-ds: Yeah, I am not sure about introducing new context as it would only react to changes in the `config` that we get from the server. Plugin activation based on the config seems to me like a thing related to the config itself. Also all the logic for enabling/disabling the plugins is contained in the `pluginManager` singleton and can be made reactive by using `usePlugin` hook (renamed, `usePluginManager` before).


I moved the initialization to the `ConfigProvider`, and let's keep it there for now.

jakubduda-dsai (approved):

PR #618, opened 2025-06-12
feat(agents): add tool calls to agent

konrad-czarnota-ds:


packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -55,6 +55,8 @@ class ToolCall(BaseModel):
     tool_name: str
     tool_arguments: dict
     model_config = ConfigDict(extra="allow")
+    tool_call_id: str
```
ds-juliamay, line 58: I think we don't need these fields to be mentioned explicitly here because they can be included in model_config = ConfigDict(extra="allow")

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -93,10 +102,47 @@ async def run(self, *args: Any, **kwargs: Any) -> AgentResult[PromptOutputT]:  #
...
+        """
+        prompt = self.prompt(input)
+        while True:
+            response = await self.llm.generate_with_metadata(
+                prompt, options=options, tools=list(self.tools_mapping.values())
+            )
+            if response.tool_calls:
+                for tool_call in response.tool_calls.tool_calls:
+                    if tool_call.tool_type == "function":
+                        tool_call_result = self.tools_mapping[tool_call.tool_name](**tool_call.tool_arguments)
```
mhordynski: what it function is async? I think we need to support that as majority of our code is asynchronous

packages/ragbits-core/src/ragbits/core/llms/base.py
ghost: Without this mypy complains

PR #623, opened 2025-06-13
feat: add tools to agents

konrad-czarnota-ds:


mhordynski:
I am wondering about the result type that we have in `Agent`...

What if we were to make it similar to the response type used in the [ChatInterface](https://ragbits.deepsense.ai/api_reference/chat/interface/chat_interface/ )

It might look something like this:

```python
agent = Agent(...)
async for response in agent.run_streaming("What is the weather in London?"):
    if text := response.as_text():
        print(f"Text: {text}")
    elif ref := response.as_tool_call():
        print(f"I am calling tool: {tool.name}")
```

Then, we'll have expandable system with type-safe methods to extract only necessary information from agents. 

 For now types would be:
 `TextResponse` / `LLMResponse` (not sure about that one)
 `ToolCallStart`
 `ToolCallEnd`

micpst:
@mhordynski i think we can just return str here for text chunks and `ToolCall`/`TooCallResult` objects for tool calls, it'd just make the interface consistent with `LLM`.

Opened a PR #627 with the new interface, need to fix tests, but its functional.

micpst (approved):

PR #627, opened 2025-06-16
review: unify agent interface

micpst:


packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -57,7 +69,6 @@ def __init__(
         prompt: type[Prompt[PromptInputT, PromptOutputT]],
         *,
         default_options: AgentOptions[LLMClientOptionsT] | None = None,
-        tools: list[Callable] | None = None,
```
mhordynski, line 60: I think tools should stay as separate parameter in __init__ - it is so common to add them in Agent interface that it would be cumbersome to always import AgentOptions.

PR #630, opened 2025-06-17
feat: support batch generation

ds-sebastianchwilczynski:


packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -5,7 +5,7 @@
 from typing_extensions import Self, TypeVar
 
 ChatFormat = list[dict[str, Any]]
-PromptOutputT = TypeVar("PromptOutputT", default=str)
+PromptOutputT_co = TypeVar("PromptOutputT_co", default=str, covariant=True)
```
mhordynski: we should not change this name without very strong reasons - this is a breaking change

ds-sebastianchwilczynski: I believe I modified it while dealing with mypy, but maybe this is not necessary, or can be solved in another way - will try to revert this change.

ds-sebastianchwilczynski: I did a revert and mypy is happy

mhordynski (approved):

PR #632, opened 2025-06-17
feat: add agent support for mcp servers

### LINKED ISSUE #613: feat: add MCP Servers as source of tools
### Feature description

Ragbits agent interface should allow to pass MCP servers as source of tools

```python
llm = LiteLLM(model_name="gpt-4o-2024-08-06", use_structured_output=True)
agent = Agent(llm=llm, prompt=AnimalPrompt, mcp_servers=[server])
await agent.run(AnimalPromptInput(animal="cat"))
```

### Motivation

MCPs allow to quickly integrate high variety of already implemented tools into a ragbits workflow

### Additional context

Depends on #568

micpst:
closes #613


packages/ragbits-agents/src/ragbits/agents/mcp/server.py
ds-sebastianchwilczynski, line 1: Do we need to implement all servers abstractions by ourselves?

[FastMCP](https://github.com/jlowin/fastmcp) has become an official part of of python MCP SDK. If you go to the [MCP SDK quickstart](https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#quickstart) You'll see they suggest to create servers using it 

image


If I'm mistaken and this is more than FastMCP let me know

ds-sebastianchwilczynski: I see these are probably just clients to be used to fetch the data from already running MCP server.

packages/ragbits-agents/src/ragbits/agents/mcp/server.py
```diff
@@ -0,0 +1,447 @@
...
+    from mcp import Tool as MCPTool
+    from mcp.client.sse import sse_client
+    from mcp.client.streamable_http import GetSessionIdCallback, streamablehttp_client
+    from mcp.shared.message import SessionMessage
+    from mcp.types import CallToolResult, InitializeResult
+
+logger = logging.getLogger(__name__)
+
+
+class MCPServer(ABC):
```
ds-sebastianchwilczynski, line 24: I believe this naming is just unfortunate, as technically shoudln't this more a MCPServerClient?

My logic is that these classes will connect with already running mcp server via `sse` or ` http`. Alternatively use Client Session, that will run server by itself?

micpst: yes, technically it's a client, but since the naming convention in other frameworks is to call it a server in the agent interface, i'd leave it as is

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -335,18 +332,27 @@ def _get_prompt_with_history(self, input: PromptInputT) -> SimplePrompt | Prompt
 
         return SimplePrompt(curr_history)
 
-    async def _execute_tool(self, tool_call: ToolCall) -> ToolCallResult:
+    async def _get_all_tools(self) -> dict[str, Tool]:
+        all_tools = [Tool.from_callable(tool) for tool in self.tools or []]
+        if self.mcp_servers:
+            all_tools.extend(await get_all_tools(self.mcp_servers))
+        return {tool.name: tool for tool in all_tools}
+
```
maxpill: What if a tool name from an MCP server conflicts with a tool from another source?

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -358,21 +364,21 @@ async def _execute_tool(self, tool_call: ToolCall) -> ToolCallResult:
             result=tool_output,
         )
 
-    def get_agent_card(
+    async def get_agent_card(
```
maxpill, line 380: Breaking change - we can mention about it in changelog

micpst: its not breaking change cuz we didn't make a release ;)

packages/ragbits-agents/src/ragbits/agents/_main.py
micpst: good catch, will raise error

akotyla:
lgtm, I tested your code and everything works great :muscle:

mhordynski (approved):

PR #641, opened 2025-06-18
feat: ui improvements

jakubduda-dsai:


typescript/ui/src/contexts/HistoryContext/HistoryContextProvider.tsx
```diff
@@ -111,6 +110,35 @@ export function HistoryProvider({ children }: PropsWithChildren) {
...
+      .find(
+        (message) =>
+          message.role === MessageRole.ASSISTANT && message.isLoading,
+      );
+
+    if (lastLoadingMessage) {
+      setMessageLoading(lastLoadingMessage.id, false);
+    }
+    chatFactory.cancel();
+  }, [history, setMessageLoading, chatFactory.cancel]);
```
dazy-ds: I think this is a little bit over complicated. We only allow one message to be streamed and we already had `isLoading` and `stopAnswering` methods, that could be used to achieve the same goal.

I also think that the loading shouldn't be a part of the message state, as in my opinion it's global characteristic of the given chat.

In my opinion we should delete these methods, remove `isLoading` from the message and use `isLoading` as it was defined before.

dazy-ds (changes requested):
Looks good, I left one major remark. Let me know what do you think about it.

PR #642, opened 2025-06-18
feat: add logfire integration

sbetlewski:


examples/core/audit/logfire_.py
```diff
@@ -0,0 +1,150 @@
...
+
+    ```bash
+    uv run examples/core/audit/logfire_.py
+    ```
+
+To visualize the metrics collected by Ragbits, follow these steps:
+
+    1. Navigate to your project in Logfire.
+    2. To check collected traces, go to the Live section in the top bar.
+    3. To check collected metrics, go to the Dashboards section and select the dashboard you created before.
```
micpst, line 28: It'd be great if you could provide a custom dashboard here, so the user can import it to test the integration. We have something similar for grafana in `config` folder

packages/ragbits-core/pyproject.toml
```diff
@@ -66,6 +66,10 @@ promptfoo = [
 otel = [
     "opentelemetry-api>=1.27.0,<2.0.0",
 ]
+logfire = [
+    "logfire>=3.19.0,<4.0.0",
```
micpst, line 70: I see that `otel` is required in order to have `logfire`, maybe lets put `otel` deps here explicitly so the user will only install one extra instead of two

packages/ragbits-core/src/ragbits/core/audit/metrics/logfire.py
```diff
@@ -0,0 +1,22 @@
...
+
+from ragbits.core.audit.metrics.otel import OtelMetricHandler
+
+
+class LogfireMetricHandler(OtelMetricHandler):
+    """
+    Logfire metric handler.
+    """
+
+    def __init__(self, provider: MeterProvider | None = None, metric_prefix: str = "ragbits") -> None:
```
micpst: Is this provider needed here? Logfire configures everything automatically afaik

packages/ragbits-core/src/ragbits/core/audit/metrics/logfire.py
```diff
@@ -0,0 +1,22 @@
...
+        """
+        Initialize the LogfireMetricHandler instance.
+
+        Args:
+            provider: The meter provider to use.
+            metric_prefix: Prefix for all metric names.
+        """
+        logfire.configure()
```
micpst: Lets maybe expose `configure` params in the initializator, as *args, **kwargs

packages/ragbits-core/src/ragbits/core/audit/metrics/logfire.py
micpst: also not sure about this `metric_prefix`, have you checked that its not overridden by logfire?

packages/ragbits-core/src/ragbits/core/audit/traces/logfire.py
```diff
@@ -0,0 +1,20 @@
...
+
+from ragbits.core.audit.traces.otel import OtelTraceHandler
+
+
+class LogfireTraceHandler(OtelTraceHandler):
+    """
+    Logfire trace handler.
+    """
+
+    def __init__(self, provider: TracerProvider | None = None) -> None:
```
micpst: ditto, do we need this provider here?

packages/ragbits-core/src/ragbits/core/audit/traces/logfire.py
```diff
@@ -0,0 +1,20 @@
...
+        """
+        Initialize the LogfireTraceHandler instance.
+
+        Args:
+            provider: The tracer provider to use.
+        """
+        logfire.configure()
```
micpst: ditto, lets expose the params

packages/ragbits-core/src/ragbits/core/audit/metrics/logfire.py
sbetlewski: provider removed as it is not needed here, `metric_prefix` is supported and not overridden

examples/core/audit/logfire_.py
sbetlewski: The expressions used to visualize histograms in Grafana are not supported by Logfire. 
I was trying to convert them into SQL queries but was unsuccessful. 
There are ongoing [discussions](https://github.com/pydantic/logfire/issues/633) about this, so Logfire may address it soon.

PR #643, opened 2025-06-18
feat: improve configurability of ui with config file on be

dazy-ds:


typescript/ui/src/App.tsx
```diff
@@ -116,6 +119,9 @@ You can ask me anything! I can provide information, answer questions, and assist
   );
 
   const content = showHistory ? historyComponent : heroComponent;
+  const logo = customization?.header?.logo ?? "🐰";
+  const title = customization?.header?.title ?? "Ragbits Chat";
+  const subTitle = customization?.header?.subtitle ?? "by deepsense.ai";
```
jakubduda-dsai: I like UI changes, one idea, can we create in UI app also some file called config.ts or so and move create there variables like logo/title/subtitle and others? so there would not be inline strings? What you think?
If you agree can you please check code for other places like this?

PR #646, opened 2025-06-23
feat: agent streaming and history 

ds-sebastianchwilczynski:


packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -57,23 +100,30 @@ class Agent(
...
 
     Current implementation is highly experimental, and the API is subject to change.
     """
 
     options_cls: type[AgentOptions] = AgentOptions
     default_module: ClassVar[ModuleType | None] = agents
     configuration_key: ClassVar[str] = "agent"
-
+    
+    #! Should we have a method to compress the history, expect compressor object here or in this method (?)
```
micpst, line 112: this is a good idea, but not feasible in the current implementation, because the compressor interface is in the `ragbits-chat` package, which cannot be a dependency for 2-layer components

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -57,23 +100,30 @@ class Agent(
...
         *,
+        history: ChatFormat | None = None,
+        keep_history: bool = False,
         tools: list[Callable] | None = None,
         default_options: AgentOptions[LLMClientOptionsT] | None = None,
     ) -> None:
+        # this is misleading - prompt here can be either a system_message, or prompt_type, it has to do nothing with a true prompt that goes to the LLM
+        # Why don't we create two separate attributes: promp_type and system_message, both optional?
+        # additionally I can get history here -  Does the history define the prompt? - Not necessarily. 
+        # But if I get the history and a prompt and have keep history set to True - I need to
```
micpst, line 126: `it has to do nothing with a true prompt that goes to the LLM` - prompt types also carry information about the prompt itself they just need to be parametrized sometimes, but they are valid prompts

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -135,61 +211,68 @@ async def run(self, *args: Any, **kwargs: Any) -> AgentResult[PromptOutputT]:
...
         merged_options = (self.default_options | options) if options else self.default_options
         llm_options = merged_options.llm_options or None
 
+        # How to handle situation if this is not the 1st interaction?
         prompt = self._create_prompt(input)
         tool_calls = []
 
+        if self.history:
+            #Ton of things can happen here.
+            #1. Someone provided a str - which is considered a system message and a history - what to do then? This combination shouldn't be supported IMO.
```
micpst, line 224: we can raise exception here, we optimize for the case when we have a chat history from the database, in which case there is no system prompt in the history

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -135,61 +211,68 @@ async def run(self, *args: Any, **kwargs: Any) -> AgentResult[PromptOutputT]:
...
-                    )
-                    prompt = prompt.add_tool_use_message(
-                        tool_call_id=tool_call.id,
-                        tool_name=tool_call.name,
-                        tool_arguments=tool_call.arguments,
-                        tool_call_result=tool_output,
-                    )
+                    # I wonder if this should be on us or on the user?
+                    # What in scenario if someone want to evaluate the tool output and do some logic based on it?
+                    # So maybe this should be an iterator too?
```
micpst, line 250: this is out of the scope of this task

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -135,61 +211,68 @@ async def run(self, *args: Any, **kwargs: Any) -> AgentResult[PromptOutputT]:
...
                 metadata=response.metadata,
                 tool_calls=tool_calls or None,
+                history=self.history,
             )
-            return outputs.result
 
     def _create_prompt(self, input: PromptInputT) -> SimplePrompt | Prompt[PromptInputT, PromptOutputT]:
+        #! Doesn't this break possibility of handling Agent History?
+        # Because we do create a new prompt instance for each run?
+        # This is just for the runs where keep_history is False.
```
micpst, line 275: yes, we need to adjust it for handling history so we won't reset the state

packages/ragbits-agents/src/ragbits/agents/_main.py
micpst: maybe is would be good idea to make it more generic and just allow for some sort of callback here, but its out of scope of this task

ds-sebastianchwilczynski:
Closing, as this will be splitted into 2 PRs

PR #647, opened 2025-06-23
feat: implement Ragbits Python chat client

maxpill:
- Added RagbitsChatClient and AsyncRagbitsChatClient for synchronous and asynchronous interactions with the Ragbits API.
- Introduced utility functions for building API URLs and parsing server-sent events.
- Defined message and response types using Pydantic models for structured data handling.
- Updated dependencies in pyproject.toml to include httpx for HTTP requests.


packages/ragbits-chat/src/ragbits/chat/client.py
```diff
@@ -0,0 +1,244 @@
+from __future__ import annotations
```
maxpill: split this file i guess, a big dumpster, too bloated

packages/ragbits-chat/src/ragbits/chat/types.py
```diff
@@ -0,0 +1,102 @@
+from __future__ import annotations
```
maxpill: split?

scripts/test_api_client.py
```diff
@@ -0,0 +1,44 @@
+# TODO: dev script for testing the Ragbits API client. To be deleted.
```
maxpill: remember to remove this script

maxpill:
Screenshot 2025-06-23 at 14 24 57

works like a charm

packages/ragbits-chat/src/ragbits/chat/client.py
```diff
@@ -0,0 +1,240 @@
...
+
```
maxpill: add error handling

packages/ragbits-chat/src/ragbits/chat/client.py
```diff
@@ -0,0 +1,240 @@
...
+from collections.abc import AsyncGenerator, Generator
```
maxpill: add interface

.gitignore
```diff
@@ -9,6 +9,9 @@ venv/
 __pycache__/
 **.egg-info/
 
+# Local cursor rules
+.cursor/rules/local/
+
```
maxpill, line 14: Had to add it. It is the only way for me to use local cursor rules without the necessity to add them to the project.

ds-sebastianchwilczynski:
Could you put new classes in the API Reference and also define how-to of setting up the client it can be a continuation of https://ragbits.deepsense.ai/how-to/chatbots/api/ - btw. this tutorial is outdated `ragbits.api` is a non-existing package rn. I know this is not part of the task but could you fix these too?

ds-sebastianchwilczynski:
image
I'd say this may be misleading as it says path and you have to ommit `.py` suffix to make it work

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+class AsyncConversation(AsyncConversationBase):
+    """Represents a single **asynchronous** chat conversation."""
+
+    def __init__(self, *, base_url: str, http_client: httpx.AsyncClient) -> None:
+        self._base_url = base_url.rstrip("/")
+        self._client = http_client
+
+        self.history: list[Message] = []
```
ds-sebastianchwilczynski, line 29: As now we do have the possibility of handling history inside the Agent class, maybe it should be handled on the back end side and this particular client shouldn't care? @mhordynski wdyt?

This can be left for another task, but I think it should be done.

Also, we should start recommending utilising Agent instead of bare run_streaming inside `chat`?

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
ds-sebastianchwilczynski, line 1: What are usecases of async version of the client/conversation?

packages/ragbits-chat/src/ragbits/chat/clients/conversation/sync_conversation.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def ask(self, message: str, *, context: dict[str, Any] | None = None) -> str:
+        """Send *message* and return the final assistant reply as a string."""
+        parts: list[str] = []
+        for chunk in self.send_message(message, context=context):
+            if chunk.type is ChatResponseType.TEXT:
+                parts.append(str(chunk.content))
+        return "".join(parts)
+
+    def send_message(
```
ds-sebastianchwilczynski: I don't get it fully:

1. `ask` method is a non-streaming one - user will not see any output until the model finishes generation
2. `send_message` seems to be one to be used if someone wants to stream the response, but then they need to handle ChatResponses by themselves

I'd say that the naming of these methods should indicate what are they for, docstring explain it, but for me it's not enough

packages/ragbits-chat/src/ragbits/chat/clients/conversation/sync_conversation.py
```diff
@@ -0,0 +1,121 @@
...
+        self.history: list[Message] = []
+        self.conversation_id: str | None = None
+        self.server_state: StateUpdate | None = None
+        self._streaming_response: httpx.Response | None = None
+
+    def ask(self, message: str, *, context: dict[str, Any] | None = None) -> str:
+        """Send *message* and return the final assistant reply as a string."""
+        parts: list[str] = []
+        for chunk in self.send_message(message, context=context):
+            if chunk.type is ChatResponseType.TEXT:
```
ds-sebastianchwilczynski, line 43: What about handling other types of responses type here, like tool calling and reference. It would be useful to see what tools were called with what arguments.

And I see that `ragbits.chat.interface` does support different static methds that returns different chat response types to it is possible to receive such

ds-sebastianchwilczynski: we can also ignore this method completely and make it on user to concatenate the message. I'd be fine with that

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
ds-sebastianchwilczynski: maybe create api endpoint that returns a history of a given conversation_id, and client just to use it?

packages/ragbits-chat/src/ragbits/chat/clients/conversation/sync_conversation.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def stop(self) -> None:
+        """Abort currently running stream (if any)."""
+        if self._streaming_response is not None and not self._streaming_response.is_closed:
+            self._streaming_response.close()
+            self._streaming_response = None
+
+    def _process_incoming(self, resp: ChatResponse, assistant_index: int) -> None:
+        """Update our local state based on *resp*."""
+        if resp.type is ChatResponseType.STATE_UPDATE and isinstance(resp.content, StateUpdate):
```
ds-sebastianchwilczynski: I believe this is not a state of a server but state of a conversation, something like

{
'checking_user_status': 'done',
'completing_the_order': 'done',
}


@mhordynski am I right?

packages/ragbits-chat/src/ragbits/chat/clients/conversation/sync_conversation.py
```diff
@@ -0,0 +1,121 @@
...
+            self.server_state = resp.content
+        elif resp.type is ChatResponseType.CONVERSATION_ID and isinstance(resp.content, str):
```
ds-sebastianchwilczynski: Also ChatResponses have this functions as_text, as_reference etc. I believe they are expected to be used here.

```py
    if text := response.as_text():
        print(f"Text: {text}")
    elif ref := response.as_tool_call():
        print(f"I am calling tool: {tool.name}")
```

packages/ragbits-chat/src/ragbits/chat/types.py
ds-sebastianchwilczynski, line 1: We have a very similar class inside `interface/types.py`. Why do this was needed?

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
ds-sebastianchwilczynski: Api is stateless and keeping the history is on the client side

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
ds-sebastianchwilczynski: Yes

packages/ragbits-chat/src/ragbits/chat/clients/conversation/sync_conversation.py
ds-sebastianchwilczynski: This will be implemented in another PR

packages/ragbits-chat/src/ragbits/chat/clients/conversation/sync_conversation.py
ds-sebastianchwilczynski: This is consistent wit TypeScript.

In general we have something like

* `run`
* `run_streaming`

Lets move to run and run_Streaming

packages/ragbits-chat/src/ragbits/chat/types.py
maxpill: removed

docs/how-to/chatbots/api.md
```diff
@@ -265,3 +268,41 @@ class MyChat(ChatInterface):
         async for chunk in self.llm.generate_streaming([*history, {"role": "user", "content": message}]):
             yield self.create_text_response(chunk)
 ```
+
+## Using the Ragbits Python Client
+
+The Ragbits ecosystem provides an official Python client that makes it easy to integrate your application with the chat API without dealing with low-level HTTP details.
+
+### Quick Start
```
ds-sebastianchwilczynski: Maybe Let's remove it?  TOC has additional entry which imo doesn't bring anything
image
 

```suggestion
```

docs/how-to/chatbots/api.md
ds-sebastianchwilczynski, line 1: Can't suggest modifications but there are still lines with `.api` broken reference on lines: 37, 217,218,219

ds-sebastianchwilczynski: Also JSONSchemaFeedbackConfig seems to be renamed to just FFeedbackConfig

docs/how-to/chatbots/api.md
```diff
@@ -265,3 +268,41 @@ class MyChat(ChatInterface):
...
+
+```python
+import asyncio
+from ragbits.chat.clients import AsyncRagbitsChatClient, RagbitsChatClient
+
+async def main() -> None:
+    async with AsyncRagbitsChatClient(base_url="http://127.0.0.1:8000") as client:
+        # One-shot request
+        reply = await client.ask("What's the weather like in Paris?")
```
ds-sebastianchwilczynski:
```suggestion
        reply = await client.run("What's the weather like in Paris?")
```

docs/how-to/chatbots/api.md
```diff
@@ -265,3 +268,41 @@ class MyChat(ChatInterface):
...
+        print(reply)
+
+        # Streaming conversation
+        conv = client.new_conversation()
+        async for chunk in conv.send_message("Give me three movie recommendations"):
```
ds-sebastianchwilczynski:
```suggestion
        async for chunk in conv.run_streaming("Give me three movie recommendations"):
```

docs/how-to/chatbots/api.md
```diff
@@ -265,3 +268,41 @@ class MyChat(ChatInterface):
...
+            if text := chunk.as_text():
+                print(text, end="")
+
+
+    sync_client = RagbitsChatClient(base_url="http://127.0.0.1:8000")
+
+    # Simple one-off request
+    response = sync_client.ask("Hello, Ragbits!")
```
ds-sebastianchwilczynski:
```suggestion
    response = sync_client.run("Hello, Ragbits!")
```

docs/how-to/chatbots/api.md
```diff
@@ -265,3 +268,41 @@ class MyChat(ChatInterface):
...
+    print(response)
+
+    # Stateful conversation
+    conversation = sync_client.new_conversation()
+    for chunk in conversation.send_message("Tell me a joke"):
```
ds-sebastianchwilczynski:
```suggestion
    for chunk in conversation.run_streaming("Tell me a joke"):
```

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
```diff
@@ -0,0 +1,114 @@
...
+        """Asynchronous version of :py:meth:`Conversation.send_message`."""
+        user_msg = Message(role=MessageRole.USER, content=message)
+        self.history.append(user_msg)
+        assistant_reply = Message(role=MessageRole.ASSISTANT, content="")
+        self.history.append(assistant_reply)
+        assistant_index = len(self.history) - 1
+
+        merged_context: dict[str, Any] = {}
+        if self.server_state is not None:
+            merged_context.update(self.server_state.state)
```
ds-sebastianchwilczynski: Aren't we doing same thing twice here with `.state` and `.model_dump()`

ds-sebastianchwilczynski: I believe model_dumpt is only necessary as we do need the signature too.

packages/ragbits-chat/src/ragbits/chat/clients/conversation/async_conversation.py
```diff
@@ -0,0 +1,114 @@
...
+class AsyncConversation(AsyncConversationBase):
+    """Represents a single **asynchronous** chat conversation."""
+
+    def __init__(self, *, base_url: str, http_client: httpx.AsyncClient) -> None:
+        self._base_url = base_url.rstrip("/")
+        self._client = http_client
+
+        self.history: list[Message] = []
+        self.conversation_id: str | None = None
+        self.server_state: StateUpdate | None = None
```
ds-sebastianchwilczynski: Can we change it to `conversation_state`, as regarding to StateUpdate docstring

image

packages/ragbits-chat/tests/unit/test_conversation.py
```diff
@@ -0,0 +1,201 @@
...
+        return _DummyStreamResponse([])
+
+    http_client = httpx.Client()
+    conv = Conversation(base_url="x", http_client=http_client)
+    conv.server_state = StateUpdate(state={"a": 1}, signature="sig")
+    conv.conversation_id = "c1"
+    with patch.object(http_client, "stream", side_effect=_mock_stream):
+        list(conv.run_streaming("question", context={"b": 2}))
+
+    print(captured["context"])
```
ds-sebastianchwilczynski:
```suggestion
```

ds-sebastianchwilczynski (approved):

packages/ragbits-chat/src/ragbits/chat/clients/__init__.py
micpst, line 1: comments to the module as a whole:
- i'd move these `client` and `conversation` client implementations to single file called `client.py` in the root of the package
- lets merge sync and async clients and add support for both sync and async methods under one namespace, i'd just suffix sync methods with `sync` since Ragbits is natively async

packages/ragbits-chat/src/ragbits/chat/clients/base.py
micpst, line 1: i'm not sure if we need these base classes here, since its a client to `ChatAPI` we don't expect to extend this type on a base level

micpst (commented):
should simplify a lot, open to discussion if needed

packages/ragbits-chat/src/ragbits/chat/clients/__init__.py
mhordynski: I'll make a small refactor later, for now merging to main

PR #648, opened 2025-06-23
feat: history handling for agents

### LINKED ISSUE #645: feat(agents): add history handling to agents
### Feature description

Currently, agents don’t have a straightforward way to manage or persist their conversation history across runs. This makes it tricky to build multi-turn, stateful conversations.

Here’s what I’d like to see for agent history handling:

1. Pass in initial history on creation

```python
my_history: ChatFormat = [{"role": "user", "content": "Hello, I'm a teapot"}]
agent = Agent(llm=..., prompt=..., history=my_history)
```

2. Get updated history after a run

```python
result = await agent.run(...)

print(result.history)  # returns the updated history

# You can use this history to spin up a new agent instance
new_agent = Agent(..., history=result.history)
```

3. Internal history can be kept automatically

```python
agent = Agent(..., keep_history=True)

await agent.run("My name is Tom")      # => hi Tom
await agent.run("Do you remember my name?")  # => of course! You are Tom
```

### Motivation

History handling is essential to keep conversation flow and short-term memory inside agents.

### Additional context

_No response_

ds-sebastianchwilczynski:
closes #645

### DIFF
--- a/docs/how-to/prompts/use_prompting.md
+++ b/docs/how-to/prompts/use_prompting.md
@@ -36,6 +36,9 @@ async def main():
 asyncio.run(main())
 ```
 
+!!! tip
+    Prompts also support history in OpenAI Chat Format, try `#!py JokePrompt([{"role": "assistant", "content": "What you call a cute door?/n A-door-able"}])`
+
 ### Extending the Prompt with an Input Model
 To extend the prompt with an input model, define a Pydantic model for the input data and pass it as a generic type to the [`Prompt`][ragbits.core.prompt.Prompt] class. The output type defaults to string.
 
--- a/docs/how-to/prompts/use_prompting.md
+++ b/docs/how-to/prompts/use_prompting.md
@@ -37,7 +37,7 @@ asyncio.run(main())
 ```
 
 !!! tip
-    Prompts also support history in OpenAI Chat Format, try `#!py JokePrompt([{"role": "assistant", "content": "What you call a cute door?/n A-door-able"}])`
+    Prompts also support history in OpenAI Chat Format, try `#!py JokePrompt(history=[{"role": "assistant", "content": "What you call a cute door?/n A-door-able"}])`
 
 ### Extending the Prompt with an Input Model
 To extend the prompt with an input model, define a Pydantic model for the input data and pass it as a generic type to the [`Prompt`][ragbits.core.prompt.Prompt] class. The output type defaults to string.
--- a/packages/ragbits-agents/src/ragbits/agents/_main.py
+++ b/packages/ragbits-agents/src/ragbits/agents/_main.py
@@ -72,7 +72,7 @@ class Agent(
     def __init__(
         self,
         llm: LLM[LLMClientOptionsT],
-        prompt: str | type[Prompt[PromptInputT, PromptOutputT]] | None = None,
+        prompt: str | type[Prompt[PromptInputT, PromptOutputT]] | Prompt[PromptInputT, PromptOutputT] | None = None,
         *,
         history: ChatFormat | None = None,
         keep_history: bool = False,
@@ -88,6 +88,7 @@ class Agent(
                 - str: A string prompt that will be used as system message when combined with string input,
                     or as the user message when no input is provided during run().
                 - type[Prompt]: A structured prompt class that will be instantiated with the input.
+                - Prompt: Already instantiated prompt instance
                 - None: No predefined prompt. The input provided to run() will be used as the complete prompt.
             history: The history of the agent.
             keep_history: Whether to keep the history of the agent.
@@ -209,7 +210,7 @@ class Agent(
                 history=updated_history.chat,
             )
 
-    def _get_updated_history(self, input: PromptInputT) -> SimplePrompt | Prompt:
+    def _get_updated_history(self, input: PromptInputT) -> SimplePrompt | Prompt[PromptInputT, PromptOutputT]:
         curr_history = deepcopy(self.history)
         if isinstance(self.prompt, type) and issubclass(self.prompt, Prompt):
             # If we had actual instance we could just run add_user_message here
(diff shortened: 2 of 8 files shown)


packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -46,6 +46,71 @@ def list_images(self) -> list[str]:  # noqa: PLR6301
...
+        Returns:
+            Prompt[PromptInputT, PromptOutputT]: The current prompt instance to allow chaining.
+        """
+        if hasattr(self, "_conversation_history"):
+            if isinstance(message, BaseModel):
+                message = message.model_dump_json()
+            self._conversation_history.append({"role": "assistant", "content": str(message)})
+            return self
+        else:
+            raise ValueError("Cannot add assistant message to a prompt that does not have a conversation history")
```
mhordynski: Maybe let's initialize new empty list in this case?

packages/ragbits-core/src/ragbits/core/prompt/base.py
```diff
@@ -46,6 +46,71 @@ def list_images(self) -> list[str]:  # noqa: PLR6301
...
+
+    def add_tool_use_message(
```
mhordynski, line 69: I think BasePrompt should also have `add_user_message`  - with just simple types (`message: str | dict[str, Any]`)  and `Prompt` should extend this logic with `PromptInputT` handling

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -175,37 +185,67 @@ async def run(self, *args: Any, **kwargs: Any) -> AgentResult[PromptOutputT]:
...
+            return AgentResult(
                 content=response.content,
                 metadata=response.metadata,
                 tool_calls=tool_calls or None,
+                history=updated_history.chat,
             )
-            return outputs.result
 
-    def _create_prompt(self, input: PromptInputT) -> SimplePrompt | Prompt[PromptInputT, PromptOutputT]:
+    def _get_updated_history(self, input: PromptInputT) -> SimplePrompt | Prompt[PromptInputT, PromptOutputT]:
```
mhordynski: nit: `_get_prompt_with_history`

docs/how-to/prompts/use_prompting.md
```diff
@@ -36,6 +36,9 @@ async def main():
 asyncio.run(main())
 ```
 
+!!! tip
+    Prompts also support history in OpenAI Chat Format, try `#!py JokePrompt([{"role": "assistant", "content": "What you call a cute door?/n A-door-able"}])`
```
micpst: it doesn't seem to be true

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -69,8 +72,10 @@ class Agent(
     def __init__(
         self,
         llm: LLM[LLMClientOptionsT],
-        prompt: str | type[Prompt[PromptInputT, PromptOutputT]] | None = None,
+        prompt: str | type[Prompt[PromptInputT, PromptOutputT]] | Prompt[PromptInputT, PromptOutputT] | None = None,
         *,
+        history: ChatFormat | None = None,
+        keep_history: bool = False,
```
micpst, line 78: let's put it in `AgentOptions`

docs/how-to/prompts/use_prompting.md
ds-sebastianchwilczynski: true, forgot history=

micpst: i mean there is no way to initialize `JokePrompt` with history even with this PR, or maybe i missed sth

ds-sebastianchwilczynski:
So there is a discussion between keep history in __init__ or in AgentOptions:
* just in __init__ - Simples solution, it's set up once during the agent initialisation and then is unchanged
* Move it to AgentOptions - then we have option to dynamically modify this behavior from the `run`, additionally this is very ragbits way to parameterise components. So it unifies the interface and also makes Agents more flexible.
* Keep in both and set to `None` in AgentOptions, and select one from __init__, if it's None in options.

I'd say I'm convinced to the 2nd option. @mhordynski wdyt?

mhordynski:
@ds-sebastianchwilczynski @micpst 

Here’s how I see what goes in options vs what should be in the class constructor:

**Class constructor:**
This is for stuff that really changes how the instance works. Think things like tools, llm, or prompt in an Agent. These are the basics - change them and you have something that works totally different.

**Options:**
This is for smaller things that don’t change much about how it works. Stuff like temperature goes here - things you might want to change for just one run. If it makes sense to change it temporarily, put it here.

**About keep_history:** imo, it doesn’t make sense to turn off history for just one run - it’s something you want on or off for the whole time the agent is running.

There are a few cases where the same parameter could be in both the constructor and options, but for that to make sense:
- It’s something people set a lot (so it’s nice to have in the constructor)
- There are good reasons to change it for just one run

A good example might be `max_tool_calls`. We don’t have this yet, but it’d help stop infinite agent loops. You’d want a default (set in the constructor), but also a way to change it for one run, like if you hit the limit and user decides to continue.

micpst:
ok, I get your point, we can leave it in initializer to keep things simple

but I'd avoid putting parameter in both the initializer and options in the future, I think it could be confusing for some users and there is no much gain of omitting `AgentOptions`

docs/how-to/prompts/use_prompting.md
ds-sebastianchwilczynski: It is possible __init__ got history as requested by @mhordynski  ` def __init__(self, input_data: PromptInputT | None = None, history: ChatFormat | None = None) -> None:`


This example will work now

mhordynski (approved):

PR #650, opened 2025-06-24
feat: support run_streaming for Agent

ds-sebastianchwilczynski:


packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -56,11 +58,56 @@ class AgentOptions(Options, Generic[LLMClientOptionsT]):
     """The options for the LLM."""
 
 
+class _AgentResultStreaming(AsyncIterator[str | ToolCall | ToolCallResult]):
```
mhordynski: let's make this class public and export it in `ragbits.agents.__init__`. 
I have a scenario where I need to wrap `Agent.run_streaming` and I want to type hint result type.

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -102,40 +151,18 @@ def __init__(
         self.history = history or []
         self.keep_history = keep_history
 
-    @overload
```
micpst, line 107: why remove these overloads?

ds-sebastianchwilczynski: I believe overloads are useful if depending on the input type the output type differs. However, in this case output type is always the same, so imo they are not needed.

micpst: this overload conditions on init param types, we have similar code in prompt class and works fine and removes ambiguity, i don't see a need to change it in this PR

ds-sebastianchwilczynski: Imo this is not a moment when you use overloads, and this for me is simply incorrect.

```py
         @overload
    async def run(
        self: "Agent[LLMClientOptionsT, None, PromptOutputT]",
        input: str,
        options: AgentOptions[LLMClientOptionsT] | None = None,
    ) -> AgentResult[PromptOutputT]: ...

    @overload
    async def run(
        self: "Agent[LLMClientOptionsT, PromptInputT, PromptOutputT]",
        input: PromptInputT,
        options: AgentOptions[LLMClientOptionsT] | None = None,
    ) -> AgentResult[PromptOutputT]: ...

    @overload
    async def run(
        self: "Agent[LLMClientOptionsT, None, PromptOutputT]",
        options: AgentOptions[LLMClientOptionsT] | None = None,
    ) -> AgentResult[PromptOutputT]: ...
```

So overload 1 and 3 has exactly the same parameterised type, and their signature changes because of how init of this class works. This overload brings no information why i don't have to provide input.

2 overload is technically incorrect as it should be 

`input: PromptInputT | None` as according to the documentation of the Prompt

>    input_data: The input data to render the prompt templates with. Must be a Pydantic model
>                 instance if the prompt has an input type defined. If None and input_type is defined,
>                 a ValueError will be raised.

and then if we correct this overload how to distinguish this and the 3rd one?

Another thing, the signature of run now is 

```py
 async def run(
        self, input: str | PromptInputT | None = None, options: AgentOptions[LLMClientOptionsT] | None = None
    )
```
as *args and **kwargs were not necessary as this function technically shouldn't accept args. If the signature of this function changes it would be pretty hard to handle it with args and kwargs.

Another thing is that a lot of people may not even use static type checkers. So dynamic attributes validation if needed is much more important than overloads, which in this case are just not helpful

ds-sebastianchwilczynski: and since I was implementing run_streaming which has exactly the same signature, I thinks this is a perfect PR to unify it.

micpst: Regarding overloads 1 and 3: while they share the same parameterized type, their difference in signature corresponds to how the Agent is constructed — specifically whether input is required at runtime. I agree this isn't strictly enforced by the type system, but from a developer experience standpoint, it's helpful to surface the expected calling patterns based on the agent config. Without the overloads, that guidance would be completely lost unless someone reads the runtime logic or docstrings carefully.

As for overload 2 — you’re right that technically it should be `PromptInputT | None`, but in practice, the method will raise a `ValueError` at runtime if `None` is passed when `input_type` is defined. To me, that makes a strong case for keeping the stricter overload: we want static checkers to flag `None` as invalid in that case rather than silently accepting it. It’s an intentional trade-off in favor of clarity and early error detection.

Re: the final signature - yes, the implementation accepts `str | PromptInputT | None`, but the overloads let us provide clearer type hints for valid combinations of inputs and options. They’re not meant to capture runtime behavior exhaustively, just guide correct usage in typical scenarios.

I agree we need strong runtime validation regardless (especially for users not using type checkers), but the overloads complement that by improving editor hints and static feedback for those who do.

I encourage you play with different prompt types in [agent example](https://github.com/deepsense-ai/ragbits/blob/main/examples/agents/tool_use.py) and see how the IDE will behave with and without overloading, frankly I prefer the former.

ds-sebastianchwilczynski: Allright the argument about IDE behavior got me

But I'll argue that 1 and 3 overload should be somehow changed. Look what happens with the IDE as I define agent like this

```py
    agent = Agent(llm=llm, tools=[get_weather])
    response = await agent.run()
```

I got a hint
image

ds-sebastianchwilczynski: What about having only such two overloads?

```py
    @overload
    async def run(
        self: "Agent[LLMClientOptionsT, None, PromptOutputT]",
        input: str | None = None,
        options: AgentOptions[LLMClientOptionsT] | None = None,
    ) -> AgentResult[PromptOutputT]: ...

    @overload
    async def run(
        self: "Agent[LLMClientOptionsT, PromptInputT, PromptOutputT]",
        input: PromptInputT,
        options: AgentOptions[LLMClientOptionsT] | None = None,
    ) -> AgentResult[PromptOutputT]: ...
```

ds-sebastianchwilczynski: and then to change Prompt overloads to such

```py
    @overload
    def __init__(
        self: "Prompt[None, PromptOutputT]", input_data: None = None, history: ChatFormat | None = None
    ) -> None: ...

    @overload
    def __init__(
        self: "Prompt[PromptInputT, PromptOutputT]", input_data: PromptInputT, history: ChatFormat | None = None
    ) -> None: ...
```

As in the previous PR I was asked to add history, again saw *args and **kwargs was a little bit worried about extending their definition in the future to even 3 attributes. So I did the same as here.

ds-sebastianchwilczynski: To support my argument not to use *args and **kwargs, take a look at 

https://ragbits.deepsense.ai/api_reference/core/prompt/#ragbits.core.prompt.Prompt

If someone decides to read this API they will probably have a hard time doing so

micpst: I like it, we can change these overloads

micpst (approved):

PR #654, opened 2025-06-25
feat: new actions with streaming response

dazy-ds:
This PR adds support for new kinds of responses:
- live updates used by agents
- followup messages

I couldn't perform the build, as there are some changes in the `@ragbits/api-client` package which hasn't been uploaded to the `npm` yet. Maybe we can build and publish this new version when the PR is accepted.

For dev purposes, you can replace usage of `@ragbits/api-client` by the previous file link.


typescript/@ragbits/api-client/src/types.ts
```diff
@@ -55,6 +57,15 @@ export interface ServerState {
     signature: string
 }
 
+export interface LiveUpdate {
+    update_id: string
+    type: 'START' | 'FINISH'
```
jakubduda-dsai: WDYT about moving it to some enum so it can be imported on FE and used possibly in some switch case or so?

dazy-ds: Done C:

mhordynski (approved):

PR #657, opened 2025-06-26
feat: support A2A protocol

akotyla:


packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -246,6 +248,66 @@ def _get_prompt_with_history(self, input: PromptInputT) -> SimplePrompt | Prompt
...
+            description=description,
+            url=f"{protocol}://{host}:{port}",
+            defaultInputModes=default_input_modes or ["text"],
+            defaultOutputModes=default_output_modes or ["text"],
+            skills=skills or [self.extract_agent_skill(f) for f in self.tools_mapping.values()],
+            capabilities=capabilities or AgentCapabilities(),
+        )
+
+    @staticmethod
+    def extract_agent_skill(func: Callable) -> AgentSkill:
```
mhordynski: let's make it private

mhordynski:
I think we should not provide `AgentOrchestrator` `RoutingPrompt` `SummarizeAgentResultsPrompt` in ragbits package code, it should be part of example.

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -246,6 +248,66 @@ def _get_prompt_with_history(self, input: PromptInputT) -> SimplePrompt | Prompt
 
         return SimplePrompt(curr_history)
 
+    def to_a2a(
```
ds-sebastianchwilczynski: I wonder if `get_agent_card` wouldn't be a better name. Both methods require what Agent to Agent is, but a2a looks very enigmatic to me.

Or maybe as a compromise get_a2a_agent_card

ds-sebastianchwilczynski:
I do agree with Mateusz that `agent_orchestrator.py` `routing_prompt.py` and `summarise_results_prompt.py` should be part of examples - They are not cleaned rn.

I believe this is a perfect example for tutorial as it really connect a lot of dots from ragbits.

Regarding how-to I'd create how to called something like `how-to: make your agent available via a2a protocol`? and just show our classic weather agent being published with server.py

ds-sebastianchwilczynski:
Regarding how the a2a protocol was used based on [example from python SDK](https://github.com/a2aproject/a2a-samples/blob/main/samples/python/hosts/multiagent/host_agent.py). They construct this orchestrator in a little bit different way, they define this `send_message` method to be a tool that LLM can use. I believe this could simplify this example a lot.

Agent will just call send_message with appropriate agent name and tool will be executed automaticallt and the results will be returned, so probably summarisation prompt will be not needed neither as Agent will handle it by itself.

ds-sebastianchwilczynski (approved):
As discussed, no new example (using tools to run a2a) is necessary, it will be written as a part of tutorial writing.

PR #658, opened 2025-06-26
feat: add tracing for chat requests

szymon-klimowicz-DS:


packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -193,9 +194,26 @@ async def _handle_chat_message(self, request: ChatMessageRequest) -> StreamingRe
...
+                    context=chat_context,
+                ) as outputs:
+                    async for chunk in RagbitsAPI._chat_response_to_sse(response_generator):
+                        if chunk.startswith("data: "):
+                            data_dict = json.loads(chunk[len("data: ") :])
+                            if data_dict.get("type") == "text":
+                                response_text += str(data_dict.get("content", ""))
+                        yield chunk
+
+                    outputs.response = response_text
```
mhordynski: Let's collect also other types of responses and store them as-is (live-updates, state-updates, references, etc.)

szymonklimowicz-dsai: I've added handling of all response types, LMK if this was what we're looking for here

PR #664, opened 2025-06-30
docs: fix local llms link

jessecampos:
Fix local llms link in docs so it goes to the right page and doesn't 404.


mhordynski:
@jessecampos thank you for spotting that and contributing! :raised_hands:

PR #668, opened 2025-06-30
fix: inline code styling

jakubduda-dsai:
After call with @micpst we found out that incline-code examples in markdown are not working correctly, so to fix this I created custom overrides for multiline and inline-code components


typescript/ui/src/core/components/ChatMessage.tsx
```diff
@@ -110,6 +110,30 @@ const ChatMessage = forwardRef<HTMLDivElement, ChatMessageProps>(
                     classNames?.content,
                   )}
                   remarkPlugins={[remarkGfm]}
+                  components={{
+                    pre: ({ node, ...props }) => (
```
dazy-ds: Let's remove unused `node` argument

typescript/ui/src/core/components/ChatMessage.tsx
```diff
@@ -110,6 +110,30 @@ const ChatMessage = forwardRef<HTMLDivElement, ChatMessageProps>(
...
+                      <pre
+                        className={cn(
+                          "max-w-full overflow-auto rounded bg-gray-200 p-2 font-mono font-normal text-gray-800",
+                          theme === Theme.DARK &&
+                            "dark:bg-gray-800 dark:text-gray-200",
```
dazy-ds: The `dark:` prefix isn't needed because we determine dark theme based on the `theme` variable. Let's remove it and just leave "bg-gray-800 text-gray-200"

typescript/ui/src/core/components/ChatMessage.tsx
```diff
@@ -110,6 +110,30 @@ const ChatMessage = forwardRef<HTMLDivElement, ChatMessageProps>(
...
+                      <code
+                        className={cn(
+                          "rounded bg-gray-200 px-2 py-1 font-mono font-normal text-gray-800",
+                          theme === Theme.DARK &&
+                            "dark:bg-gray-800 dark:text-gray-200",
+                        )}
+                      >
+                        {props.children}
+                      </code>
+                    ),
```
dazy-ds: Same as above, let's remove `node` and `dark:` prefix

typescript/ui/src/core/components/ChatMessage.tsx
jakubduda-dsai: i switched simply to `{children}` since only children is used

PR #670, opened 2025-07-01
docs: how-to define and use agents

akotyla:


ds-sebastianchwilczynski (commented):
If I see correctly code content is almost copied from the example, maybe we use code blocks embedding functionality of material for mkdocs? https://squidfunk.github.io/mkdocs-material/reference/code-blocks/?h=code+blo#embedding-external-files

PR #681, opened 2025-07-01
feat: typescript docs update

dazy-ds:


typescript/@ragbits/api-client/src/types.ts
```diff
@@ -191,97 +191,70 @@ export interface StreamCallbacks<T, E = Error> {
...
+    Endpoints extends { [K in keyof Endpoints]: EndpointDefinition },
+> = Endpoints[URL]['method']
 
 /**
  * Generic request options for API endpoints with typed methods and body
  */
-export interface TypedApiRequestOptions<T extends ApiEndpointPath> {
-    method?: ApiEndpointMethod<T>
-    body?: ApiEndpointRequest<T> extends never
+export interface TypedRequestOptions<
```
jakubduda-dsai: Since we make changes in naming of types/interfaces I'd suggest to rename this interface to RequestOptions

typescript/ui/README.md
```diff
@@ -1,50 +1,144 @@
...
+
+### Example Plugin Structure
+
+```
+plugins/
+└── MyPlugin/
+    ├── index.ts       # Plugin entry point
+    ├── components/    # Plugin-specific components
+    └── types.ts       # Plugin type definitions
+```
```
jakubduda-dsai, line 100: I feel like current plugins dont match this structure, maybe it should be organized like this

typescript/@ragbits/api-client/src/types.ts
dazy-ds: Good idea, changed.

typescript/ui/README.md
dazy-ds: This is just an example structure of a plugin, but sure, we can follow it and set the convention C:
Changed.

Issue #637, opened 2025-06-18
feat: unit tests for Ragbits API Clients


jakubduda-dsai:
Implement unit tests for Ragbits API Clients

PR #684, opened 2025-07-02
feat: #637 Live updates supports markdown

jakubduda-dsai:


typescript/ui/src/core/components/LiveUpdates.tsx
```diff
@@ -68,14 +86,30 @@ export default function LiveUpdates({
...
+              <Markdown
+                className={cn(
+                  "markdown-container prose max-w-full text-sm",
+                  theme === Theme.DARK && "dark:prose-invert",
+                  classNames?.liveUpdates,
+                )}
+                remarkPlugins={[remarkGfm]}
+              >
+                {lastUpdate.description}
+              </Markdown>
```
dazy-ds: Current implementation breaks the shimmer effect for description. Make sure to add `text-transparent` as class to this `Markdown` element.

Side note: Maybe let's add a comment to the `ShimmerText` component that says that children should have `text-transparent` set for the effect to be applied correctly.

dazy-ds (changes requested):
One minor change. Also let's apply what we have talked about on Slack C:

dazy-ds (approved):
LGTM

PR #686, opened 2025-07-02
feat:  Google Drive Source 

nikpocuca:
## Additions
Adding in a google drive as a source, simply use the drive-id hash as a source identifier.  

## google-side requirements
- requires google service account
- google account json key 
- drive-id
- api-enabled

## google-side pip packages introduced. 
- google-api-python-client
- google-auth
- google-auth-httplib2


docs/how-to/sources/load-dataset.md
```diff
@@ -10,6 +10,7 @@ This is the list of currently supported sources by Ragbits.
 |-|-|-|
 | Azure Blob Storage | `azure://https://<account-name>.blob.core.windows.net/<container-name>/<blob-name>` | [`AzureBlobStorageSource`][ragbits.core.sources.AzureBlobStorageSource] |
 | Google Cloud Storage | `gcs://<bucket-name>/<prefix>` | [`GCSSource`][ragbits.core.sources.GCSSource] |
+| Google Drive | <drive-id> | ragbits.core.sources.GoogleDriveSource |
```
akotyla:
```suggestion
| Google Drive | `<drive-id>` | [`GoogleDriveSource`][ragbits.core.sources.GoogleDriveSource] |
```

akotyla: previously the URI schema was rendered to an empty string

packages/ragbits-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Added GoogleDrive support (#686)
```
akotyla:
```suggestion
- Added Google Drive support (#686)
```

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,459 @@
...
+    "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet": ".xlsx",
+    "application/vnd.openxmlformats-officedocument.presentationml.presentation": ".pptx",
+    "image/png": ".png",
+    "application/pdf": ".pdf",
+    "text/html": ".html",
+    "text/plain": ".txt",
+    "application/json": ".json",
+}
+
+LRO_EXPORT_SIZE_THRESHOLD = 9 * 1024 * 1024  # 9MB
```
akotyla: maybe we can mark these module-level constants as private? like we did here: https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-core/src/ragbits/core/sources/git.py

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,459 @@
...
+
+                    fh = io.BytesIO()
+                    downloader = MediaIoBaseDownload(fh, request)
+                    done = False
+                    while not done:
+                        _, done = downloader.next_chunk()
+
+                    fh.seek(0)
+                    with open(path, mode="wb") as file_object:
+                        file_object.write(fh.read())
```
akotyla: I'm wondering if we could write it directly to a file instead of loading it into memory, I'm thinking of something like this:
```python
with open(path, 'wb') as fh:
    downloader = MediaIoBaseDownload(fh, request)
    done = False
    while not done:
        _, done = downloader.next_chunk()
```

what do you think? :)

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
akotyla: where is this variable used in the code?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,459 @@
...
+            raise ValueError(f"Cannot directly fetch a folder. Use list_sources and iterate. Folder ID: {self.file_id}")
+
+        local_dir = get_local_storage_dir()
+        file_local_dir = local_dir / self.file_id
+        file_local_dir.mkdir(parents=True, exist_ok=True)
+
+        file_extension = ""
+        export_mime_type = self.mime_type
+
+        if self.mime_type.startswith("application/vnd.google-apps"):
```
akotyla: maybe we can extract this logic into a helper method like `_determine_file_extension`? I think this will simplify the `fetch` method a bit

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,459 @@
...
+                                "mimeType": item["mimeType"],
+                                "is_folder": False,
+                                "path_in_drive": full_local_name,
+                            }
+
+                    page_token = results.get("nextPageToken", None)
+                    if not page_token:
+                        break
+                except HttpError as e:
+                    print(f"Error listing folder {current_folder_id} (path: {current_path_prefix}): {e}")
```
akotyla: maybe it's better to use `audit.traces` or a logger instead of prints?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: its an artifact from a previous rendition, technically the google drive api itself limits the size downloaded. I will remove it for now because there isn't really a way to check without attempting to download it first.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: just updated.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: ive added the _prefix.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: added traces.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: extracted into private method.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,486 @@
...
+                cls._google_drive_client = build("drive", "v3", credentials=creds)
+                cls._google_drive_client.files().list(
+                    pageSize=1, fields="files(id)", supportsAllDrives=True, includeItemsFromAllDrives=True
+                ).execute()
+            except exceptions.GoogleAuthError as e:
+                raise ValueError("Google Drive credentials are missing or incomplete. Please configure them.") from e
+            except HttpError as e:
+                error_content = e.content.decode() if hasattr(e, "content") else ""
+                lower_error = error_content.lower()
+                if "drive api" in lower_error and ("not enabled" in lower_error or "not been used" in lower_error):
```
maxpill: How about we split this method a bit?

maxpill: maybe `_get_client` and `_initialize_client`?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,486 @@
...
+                    break
+                except Exception as e:
+                    with trace("folder_listing_unexpected_error") as outputs:
+                        outputs.error = f"An unexpected error occurred while listing folder {current_folder_id}: {e}"
+                    break
+
+        await _recursive_list(drive_id, current_root_shared_drive_id=(drive_id if is_shared_drive else None))
+
+    @classmethod
+    async def _check_drive_type(cls, client: "GoogleAPIResource", drive_id: str) -> tuple[bool, bool, str]:
```
maxpill, line 345: no need async i think

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -0,0 +1,486 @@
...
+        if len(parts) == 1:
+            client = cls._get_client()
+            try:
+                file_meta = (
+                    client.files().get(fileId=drive_id, fields="id, name, mimeType", supportsAllDrives=True).execute()
+                )
+                is_folder = file_meta["mimeType"] == "application/vnd.google-apps.folder"
+                return [
+                    cls(
+                        file_id=file_meta["id"],
```
maxpill: How about we split?

maxpill: maybe like

`@classmethod`  
  `@traceable`  
  `async def from_uri(cls, path: str) -> Iterable[Self]:`
  
  and some helpers

`@classmethod`  
  `async def _handle_single_id(cls, drive_id: str) -> Iterable[Self]:`

`@classmethod`  
  `async def _handle_path_pattern(cls, parts: list[str]) -> Iterable[Self]:`

`@classmethod`  
  `async def _handle_recursive_pattern(cls, folder_id: str) -> Iterable[Self]:`

`@classmethod`  
  `async def _handle_direct_children_pattern(cls, folder_id: str) -> Iterable[Self]:`

`@classmethod`  
  `async def _handle_prefix_pattern(cls, folder_id: str, prefix: str) -> Iterable[Self]:`

nikpocuca: Hey I am only re-factoring the first check because the other helpers would be less than 5 lines of code.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: i split the inside method only.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
nikpocuca: the execute() call can be blocking, it needs to be async.

PR #689, opened 2025-07-03
feat: debug mode

dazy-ds:


packages/ragbits-chat/src/ragbits/chat/interface/_interface.py
```diff
@@ -97,6 +97,7 @@ class ChatInterface(ABC):
     feedback_config: FeedbackConfig = FeedbackConfig()
     ui_customization: UICustomization | None = None
     history_persistence: HistoryPersistenceStrategy | None = None
+    debug_mode: bool = False
```
mhordynski: This variable probably shouldn't be on class level - I would rather expect it to be available in `__init__` and passed from `ragbits api run` command tool.

mhordynski: something like `ragbits api run --debug src.MyChat`

jakubduda-dsai: Isn't it part of the config class like the other things which come with `/api/config` endpoint?

dazy-ds: I understand Mateusz's point of view as having debug as the run argument is easier when switching between the modes. We can just write `--debug` and we are done, without the need to look for the implementation and change it there.

Personally I am fine with either of the approaches

dazy-ds: I moved the `debug_mode` to `RagbitsAPI` and made it command argument

PR #691, opened 2025-07-03
chore: run examples pipeline on release branches

mhordynski:


.github/workflows/run-examples.yml
micpst, line 1: nit: maybe lets rename this file to `release-checks.yml`, we can add more things here in the future. it'll be also consistent with rest of the files in the config

.github/workflows/run-examples.yml
```diff
@@ -0,0 +1,78 @@
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    permissions:
+      checks: write
+      pull-requests: write
+      contents: read
+
+    services:
+      postgres:
+        image: pgvector/pgvector:0.8.0-pg17
```
maxpill: Extract to vars?

`PGVECTOR_IMAGE_TAG: 0.8.0-pg17`

.github/workflows/run-examples.yml
```diff
@@ -0,0 +1,78 @@
...
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+
+      weaviate:
+        image: cr.weaviate.io/semitechnologies/weaviate:1.30.6
```
maxpill: Extract to vars?

`WEAVIATE_IMAGE_TAG: 1.30.6`

.github/workflows/run-examples.yml
```diff
@@ -0,0 +1,78 @@
+name: Check examples
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - 'release/**'
+
```
maxpill, line 8: We can add

```
concurrency:
  group: ${{ github.workflow }}-${{ github.ref }}
  cancel-in-progress: true
```

to cancel parallel builds and only run the latest

.github/workflows/run-examples.yml
mhordynski: Maybe then we can have in the future and call this file from there. It is sepearate like that to allow it to be ran manually for any branch

mhordynski: Maybe then we can have in the future and call this file from there. It is sepearate like that to allow it to be ran manually for any branch

mhordynski:
@maxpill good suggestions, applied all

PR #692, opened 2025-07-04
feat: Chat options modal form

jakubduda-dsai:


packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -179,6 +181,9 @@ async def config() -> JSONResponse:
                 "customization": self.chat_interface.ui_customization.model_dump()
                 if self.chat_interface.ui_customization
                 else None,
+                "chat": {
+                    "form": chat_config
+                },
```
dazy-ds: I am not really a fan of this nesting. Maybe rename `chat` to `chat_options` and remove nested `form`?
WDYT?

typescript/ui/src/App.tsx
```diff
@@ -84,8 +84,8 @@ export default function App() {
     setShouldAutoScroll(true);
   }, []);
 
-  const handleSubmit = (text: string) => {
-    sendMessage(text);
+  const handleSubmit = (text: string, options?: Record<string, any>) => {
```
dazy-ds: We can change the typing to `Record<string, unknown>` as we don't care about the specific types

packages/ragbits-chat/src/ragbits/chat/api.py
jakubduda-dsai: So I did it on purpose, because probably soon we will develop some functionalities like accepting images/pdfs or so and wanted to leave this object chat for extending chat options wchich are not connected to form. WDYT?

dazy-ds: That's fair. Let's keep it as it is now C:

examples/chat/chat.py
```diff
@@ -66,6 +74,7 @@ class MyChat(ChatInterface):
         dislike_enabled=True,
         dislike_form=DislikeFormExample,
     )
+    chat_config = ChatConfig(chat_form=ChatFormExample)
```
mhordynski: let's name it `UserSettings`

jakubduda-dsai: Ok, but I'd suggest to leave in `/api/config` structure like chat.user_settings/user_settings_form since prbbly in future we are going to extend chat functionality with image/files upload and that would be good place to have setters here like chat.allow_images. WDYT?

mhordynski: I think `user_settings.form` would make more sense tbh

mhordynski (approved):

PR #693, opened 2025-07-04
feat: add PPTX document parser 

maxpill:
- Introduced PptxDocumentParser for extracting content from PPTX files.
- Updated pyproject.toml to include python-pptx as a dependency.
- Modified the router to use PptxDocumentParser for PPTX document types.
- Cleaned up formatting in pyproject.toml and init.py for consistency.


examples/document-search/test_pptx_parser.py
```diff
@@ -0,0 +1,102 @@
+# This is a temporary script for development purposes and PR testing.
```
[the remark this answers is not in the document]
nikpocuca: remove temporary comment? idk if you wanted to commit this.

examples/document-search/test_pptx_parser.py
```diff
@@ -0,0 +1,102 @@
...
+# It will be removed before merging.
```
nikpocuca: it will be removed before merging.

examples/document-search/test_pptx_parser.py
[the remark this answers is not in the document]
maxpill: Yeah wanted to commit this for review purposes, so it is easier to test. Definitely needs to be removed before merging. Just ignore this file

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/pptx/extractors/extractors.py
```diff
@@ -0,0 +1,274 @@
...
+                    if content and content.strip():
+                        element = self._create_text_element(
+                            element_type=element_type,
+                            document_meta=document_meta,
+                            content=content,
+                            slide_idx=slide_idx,
+                            shape=shape,
+                        )
+                        elements.append(element)
+                except (AttributeError, TypeError):
```
nikpocuca: so the behavior here is to skip slides that fail extraction? 

Is there any logging for this?

examples/document-search/test_pptx_parser.py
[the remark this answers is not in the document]
nikpocuca: thats okay, i do the same thing

maxpill:
merge #733 first

pocucan-ds (approved):
looks good to me.

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/pptx/extractors/extractors.py
```diff
@@ -0,0 +1,329 @@
...
+
+from ragbits.document_search.documents.document import DocumentMeta
+from ragbits.document_search.documents.element import Element, ElementLocation, ImageElement, TextElement
+
+
+class BasePptxExtractor(ABC):
+    """Base class for all PPTX content extractors."""
+
+    @staticmethod
+    def _get_slides(presentation: Presentation, slide: Slide | None = None) -> list[tuple[int, Slide]]:
```
ds-sebastianchwilczynski: why as we give `Presentation` and `slide` we get only slide as a result?

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/pptx/extractors/extractors.py
```diff
@@ -0,0 +1,329 @@
...
+        return elements
+
+    @abstractmethod
+    def extract(
+        self, presentation: Presentation, document_meta: DocumentMeta, slide: Slide | None = None
+    ) -> list[Element]:
+        """Extract content from the presentation or specific slide."""
+
+    @abstractmethod
+    def get_extractor_name(self) -> str:
```
ds-sebastianchwilczynski: This is non-pythonic please use a read-only property for this

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/__init__.py
```diff
@@ -1,4 +1,11 @@
 from ragbits.document_search.ingestion.parsers.base import DocumentParser, ImageDocumentParser, TextDocumentParser
+from ragbits.document_search.ingestion.parsers.pptx import PptxDocumentParser
```
ds-sebastianchwilczynski: docling and unstructured parsers are not listed here. I'd keep it consistent and either link all or none.

Were they not listed due to import time optimisation @mhordynski?

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/pptx/parser.py
ds-sebastianchwilczynski, line 1: Both docling and unstructured in theory supports pptx too. So if this one is better we should write why and mention this in documentation too.

ds-sebastianchwilczynski (changes requested):
One big problem I see is that we create `Element` objects with extractors, and if I understand correctly these `Elements` are ment to be transformed with an Embedder as they has `to_vector_db_entry`. But in this case objects you extract should first go through grouping and chunking process. If we take every even the smallest object from pptx and embed it we will end up with ton of useless chunks and our RAG system will fail.
 
 -----
 
I tested unstrutured and docling for this task and unstructured `partition_pptx` didn't extract images from my slides. However, docling did great job.

```py
from docling.document_converter import DocumentConverter
from docling_core.types.doc.document import PictureItem
converter = DocumentConverter()
doc = converter.convert(source="../testing_scripts/Slides.pptx").document

for item, idx in doc.iterate_items(with_groups=True):
    if not isinstance(item, PictureItem):
        print(item)
        
for idx, i in enumerate(doc.pictures):
    img = i.image.pil_image
    img.save(f"Slides_{idx}.png")
```

it loooks like it didn't extract presenter's notes, and some other stuff that we did.

But after we have it extracted with this form, we have docling grouping, serialization, chunking etc for free.

Also we can add more elements to the [Docling's Document class](https://docling-project.github.io/docling/reference/docling_document/#docling_core.types.doc.DoclingDocument) 

So what, I'd do:
1. Utilise everything we can from docling directly - less code, less problems
2. Then expand docling document with our custom implementation, our implementations should return docling compliant formats. What I'd personally do is implement callback system in the Docling parser. So you can add callbacks that will be run as we are parsing a pptx file. This callback takes something, and docling document, extract value from something and adds it to docling document
3. Then I believe we can utilise Chunker from docling, or create a custom one.

But please discuss this with @mhordynski

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/__init__.py
maxpill: implemented

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/pptx/extractors/extractors.py
maxpill: implemented

packages/ragbits-document-search/src/ragbits/document_search/ingestion/parsers/pptx/extractors/extractors.py
maxpill: implemented

PR #697, opened 2025-07-07
feat(ui): add integration tests to UI package

dazy-ds:


package.json
```diff
@@ -14,6 +14,7 @@
     "dev": "npm run dev --workspaces --if-present",
     "test": "npm run test --workspaces --if-present",
     "lint": "npm run lint --workspaces --if-present",
-    "format": "npm run format --workspaces --if-present"
+    "format": "npm run format --workspaces --if-present",
+    "test:integration": "npm run test:integration --workspaces --if-present"
```
jakubduda-dsai: I now it doesnt matter but can we move this under the line with command `test`?

typescript/ui/package.json
```diff
@@ -5,11 +5,14 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
+    "dev:backend": "cd ../../ && uv sync && ragbits api run examples.chat.offline_chat:MyChat --debug",
```
jakubduda-dsai: I think that maybe path to the chat module should be passed by command line as. argument, WDYT?

package.json
dazy-ds: Sure, moved!

typescript/ui/package.json
dazy-ds: Good idea, this will make `dev:backend` more reusable.

Changed!

typescript/ui/src/main.tsx
```diff
@@ -37,9 +37,9 @@ createRoot(document.getElementById("root")!).render(
       <ThemeContextProvider>
         <RagbitsProvider baseUrl={API_URL}>
```
jakubduda-dsai: Since we change naming to HistoryContextProvider in this PR maybe we would make a change to Ragbits package and changenaming of RagbitsProvider? WDYT?

dazy-ds: Changed

PR #700, opened 2025-07-07
feat: add token usage to LLM interface

mhordynski:


packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -194,9 +194,11 @@ async def _call(
             results["logprobs"] = response.choices[0].logprobs["content"]  # type: ignore
 
         if response.usage:  # type: ignore
-            results["completion_tokens"] = response.usage.completion_tokens  # type: ignore
-            results["prompt_tokens"] = response.usage.prompt_tokens  # type: ignore
-            results["total_tokens"] = response.usage.total_tokens  # type: ignore
+            results["usage"] = {
```
ds-sebastianchwilczynski: What is the reason of not creating the `Usage` object here and in 4 other places?

packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -53,6 +53,18 @@ def parse_tool_arguments(cls, tool_arguments: str) -> dict:
         return pased_arguments
 
 
+class Usage(BaseModel):
```
mhordynski, line 58: Let's add:

- `__add__` 
- `__iadd__`
- `__str__`
- `__repr__`

ds-sebastianchwilczynski: I just added __repr__ as I belive str and repr in thsi case can be the same

PR #703, opened 2025-07-08
chore: move a2a to extra dependencies

micpst:


packages/ragbits-agents/src/ragbits/agents/_main.py
ds-sebastianchwilczynski, line 1: Could you add exceptions to public methods that utilise a2a indicating they need to install ragbits-agents[a2a]?

packages/ragbits-agents/src/ragbits/agents/a2a/server.py
```diff
@@ -1,12 +1,15 @@
...
 from urllib.parse import urlparse
 
-import uvicorn
-from a2a.types import AgentCard
-from fastapi import FastAPI
 from pydantic import BaseModel
 
 from ragbits.agents import Agent, AgentResult
 
+with suppress(ImportError):
```
ds-sebastianchwilczynski: Why are we supressing here? Aren't we supposedto raise this error, and prompt user to install ragbits-agents[a2a]?

PR #704, opened 2025-07-08
docs: Update README

akotyla:


README.md
ds-sebastianchwilczynski, line 1: I wonder if we should modify these quickstarts examples to show `Agent` class as the one which should be used by default. Because now we show llm as such, and there is no agentic example in the quickstart. 

I stand for not adding another example but making Agent a default option instead of llm.generate.

micpst: I'd leave those `LLM` classes in quickstarts, because we present there an interface for working with LLMs and building RAGs, for such use cases its's preferred to use the raw `LLM` interface, so there is no need to change that.

micpst: but we can add one more quickstart with Agentic RAG, its like ~20 lines of code with ragbits, and we can use `Agent` interface there.

micpst: I have some snippets, can add it later

README.md
```diff
@@ -19,6 +19,12 @@
 
 ## Features
 
+### 🤖 Build Multi-Agent Workflows with Ease
```
micpst: nit: would put this a bit lower, between `### 📚 Fast & Flexible RAG Processing` and `### 🚀 Deploy & Monitor with Confidence` as agents aren't "core" feature set of this repo

micpst: please update also `index.md` in `docs` folder

micpst (approved):

PR #705, opened 2025-07-08
feat: add max_turns in AgentOptions

ds-sebastianchwilczynski:


packages/ragbits-agents/tests/unit/test_agent.py
```diff
@@ -439,3 +464,22 @@ async def test_agent_history_with_tools(llm_with_tool_call: MockLLM, method: Cal
...
+    )
+
+    options: AgentOptions = AgentOptions(max_tool_calls=1)
+
+    with pytest.raises(AgentMaxToolCallsExceededError) as exc_info:
+        await method(agent, options=options)
+
+    assert exc_info.value.max_tool_calls == 1
+    assert exc_info.value.actual_tool_calls == 1
+    assert "Maximum number of tool calls exceeded: 1 > 1" in str(exc_info.value)
```
mhordynski: This message seems counterintuitive: `1 > 1` looks like it failed during the 1st tool call, but it is the 2nd

packages/ragbits-agents/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Support history handling in Agent (#648)
 - Support run_streaming in Agent (#650)
 - Support A2A protocol (#649)
+- Add Max Tools Calls to agentsptions (#705)
```
mhordynski:
```suggestion
- Add max tool calls limit to AgentOptions (#705)
```

mhordynski:
We should have a sensible default to protect our users from infinite agent loops (let's say 15 tool calls)

packages/ragbits-agents/tests/unit/test_agent.py
mhordynski: Also I would say explicitly in error how to increase this limit, for example:

> To change this limit, pass ragbits.agents.AgentOptions with max_tool_calls option when initializing the Agent.
> agent = Agent(options=AgentOptions(max_tool_calls=XX))

mhordynski (dismissed):

micpst:
@mhordynski maybe we shouldn't constrain this at the tool call level, but just limit the number of loop runs explicitly. In openai sdk there's parameter called `max_turns`, `max_tool_calls` can be harder to tune

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -293,9 +307,13 @@ async def _stream_internal(
                         yield result
                         prompt_with_history = prompt_with_history.add_tool_use_message(**result.__dict__)
                         returned_tool_call = True
+                        turn_count += 1
```
micpst: this param seems broken for streaming mode, it looks like we still count tool calls not turns here. i didn't test it, pls verify if i'm wrong

ds-sebastianchwilczynski: good catch, I thought tests would differentiate between this cases, but with just one tool call these are the same. Will add another thest with many tools calls but just 2 turns

PR #706, opened 2025-07-09
refactor(ui): external history store

dazy-ds:


typescript/ui/src/core/components/PromptInput/PromptInput.tsx
```diff
@@ -21,8 +21,8 @@ import { MessageRole } from "@ragbits/api-client-react";
 import PluginWrapper from "../../utils/plugins/PluginWrapper";
 import { ChatOptionsPlugin } from "../../../plugins/ChatOptionsPlugin";
 
-interface PromptInputProps {
-  submit: (text: string, options?: Record<string, unknown>) => void;
+export interface PromptInputProps {
+  submit: (text: string) => void;
   stopAnswering: () => void;
   onArrowUp?: (isFirstLine: boolean) => void;
   onArrowDown?: (isLastLine: boolean) => void;
```
jakubduda-dsai: so since we export and use this interface in QuickMessageInput maybe we should make some structure like `components/inputs/types.ts` and there store types and store inputs components in `components/inputs/[PromptInput][QuickMessageInput]`. WDYT?

typescript/ui/src/core/components/ChatMessage.tsx
```diff
@@ -18,20 +18,23 @@ export type ChatMessageProps = {
...
+    const message = useMessage(messageId);
+
+    if (!message) {
+      throw new Error("Tried to render non-existent message");
+    }
+
+    const { serverId, content, role, references, liveUpdates } = message;
     const rightAlign = role === MessageRole.USER;
+    const isLoading =
+      isHistoryLoading && role === "assistant" && messageId === lastMessageId;
```
jakubduda-dsai: can we use here MessageRole.ASSISTANS instead of string?

typescript/ui/src/core/components/PromptInput/PromptInput.tsx
dazy-ds: I removed the `export` and used `ComponentProps<typeof PromptInput>` instead. I would keep the props types collocated until we really have them used across the app.

I applied your suggestion to group inputs in one directory C:

typescript/ui/src/core/components/ChatMessage.tsx
dazy-ds: Good catch, thanks!

PR #708, opened 2025-07-09
refactor: simplify ragbits-chat client

mhordynski:


packages/ragbits-chat/src/ragbits/chat/client/client.py
```diff
@@ -0,0 +1,132 @@
...
+        tb: TracebackType | None,
+    ) -> None:
+        """Ensure the underlying async HTTP session is closed on exit."""
+        await self.aclose()
+
+
+class SyncRagbitsChatClient:
+    """Stateless *synchronous* Ragbits chat client.
+
+    The sole responsibility of this class is to spawn new
```
ds-sebastianchwilczynski, line 78: nit: Asynchronous Client doesn't have this part of the docs

PR #716, opened 2025-07-14
feat: new API for attachments in prompts

akotyla:


packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -286,14 +294,24 @@ def list_images(self) -> list[str]:
...
+            if not detected or not detected.mime.startswith("image/"):
                 raise PromptWithImagesOfInvalidFormat()
-            image_url = f"data:{detected_type.mime};base64,{base64.b64encode(image).decode('utf-8')}"
+            return f"data:{detected.mime};base64,{base64.b64encode(img_bytes).decode('utf-8')}"
+
+        if isinstance(image, Attachment):
+            if isinstance(image.data, bytes):
+                image_url = _encode_image_bytes_to_url(image.data)
+            elif image.url is not None:
+                return {"type": "image_url", "image_url": {"url": image.url, "format": f"image/{image.mime_type}"}}
```
mhordynski: Not always image, it can be pdf etc

packages/ragbits-core/tests/unit/prompts/test_prompt.py
```diff
@@ -49,6 +50,19 @@ def _get_image_bytes() -> bytes:
         return f.read()
 
 
+def _get_pdf_bytes() -> bytes:
```
ghost: This doesn't seem to be used anywhere

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -18,6 +19,14 @@
 FewShotExample = tuple[str | PromptInputT, str | PromptOutputT]
 
 
+class Attachment(BaseModel):
+    """Represents an image attachment that can be passed to a LLM."""
```
ghost: Does it only support image?

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -80,6 +89,13 @@ def _render_template(cls, template: Template, input_data: PromptInputT | None) -
             context = input_data.model_dump(serialize_as_any=True)
         return template.render(**context)
 
+    @classmethod
+    def _get_files_from_input_data(cls, input_data: PromptInputT | None | str) -> list[Attachment]:
```
ghost: This doesn't seem to be used anywhere

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -158,6 +187,7 @@ def __init__(self, input_data: PromptInputT | None = None, history: ChatFormat |
         )
         self.rendered_user_prompt = self._render_template(self.user_prompt_template, input_data)
         self.images = self._get_images_from_input_data(input_data)
```
ghost: Shouldn't attachements replace images?

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
```diff
@@ -42,6 +56,12 @@ class Prompt(Generic[PromptInputT, PromptOutputT], BasePromptWithParser[PromptOu
     output_type: type[PromptOutputT]
     system_prompt_template: Template | None
     user_prompt_template: Template
+
+    warnings.warn(
```
ds-sebastianchwilczynski: This should be displayed only if image_input_fields is set to something different than None. Now even if I use Attachments this is displayed

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
ds-sebastianchwilczynski: This is probably not only image attachment

packages/ragbits-core/tests/unit/prompts/test_prompt.py
```diff
@@ -122,82 +151,158 @@ class TestPrompt(Prompt[_PromptInput, str]):  # pylint: disable=unused-variable
...
+    class TestAttachmentPrompt(Prompt):
+        user_prompt = "What is in this PDF?"
+
+    prompt = TestAttachmentPrompt(_SingleAttachmentPromptInput(attachment=_get_pdf_bytes()))
+    pdf_list = prompt.list_pdfs()
+    assert len(pdf_list) == 1
+    assert pdf_list[0].startswith("data:application/pdf")
+
+
+@pytest.mark.parametrize(
```
ds-sebastianchwilczynski, line 287: I believe this test_pdf and test_pdfs tests could be just one test. Same with images

docs/how-to/prompts/use_attachments_in_prompts.md
```diff
@@ -0,0 +1,159 @@
...
+    photo = Attachment(url="<your_photo_here>")
+    prompt = AnimalPhotoPrompt(AnimalPhotoInput(photo=photo))
+    response = await llm.generate(prompt)
+    print(response)
+
+
+asyncio.run(main())
+```
+
+### Using multiple attachments as input
```
ds-sebastianchwilczynski: imo this example would be enough and the one above can be removed

examples/core/prompt/multimodal.py
```diff
@@ -25,6 +25,7 @@
 
 from ragbits.core.llms import LiteLLM
 from ragbits.core.prompt import Prompt
+from ragbits.core.prompt.prompt import Attachment
```
ds-sebastianchwilczynski: `prompt.prompt` looks a little bit weird. Maybe we move Attachment to __init__ and __all__?

examples/core/prompt/multimodal.py
ds-sebastianchwilczynski, line 1: Maybe we add a pdf example too, so that we have it tested.

examples/core/prompt/multimodal_with_few_shots.py
```diff
@@ -25,6 +25,7 @@
 
 from ragbits.core.llms import LiteLLM
 from ragbits.core.prompt import Prompt
+from ragbits.core.prompt.prompt import Attachment
```
ds-sebastianchwilczynski: same as above

ds-sebastianchwilczynski (commented):
it all looks good, I run it and it works. Suggested changes are not necessary,let me know wdyt.

examples/core/prompt/multimodal.py
sbetlewski: great catch, thanks for this!

packages/ragbits-core/tests/unit/prompts/test_prompt.py
sbetlewski: I don't think so... 
we use different prompt inputs for each of them: `_SingleAttachmentPromptInput` and `_MultipleAttachmentPromptInput`; thanks to this separation, we can test both single `Attachment` object and the list of them

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
sbetlewski: Unfortunately, yes. The difficulty here is that there is nothing like `@deprecated` annotation for class attributes in Python.
I was trying to implement the `__get__` and `__set__` methods for the `image_input_fields` attribute and to warn user when it's accessed, but since this is a class attribute, it is overriden each time a new subclass is created - so this approach does not work.
Do you have any other ideas to work around it?

examples/core/prompt/multimodal.py
sbetlewski: good idea, done

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
ds-sebastianchwilczynski: you could check the value in __init__? Probably also not the best, but will work.

packages/ragbits-core/tests/unit/prompts/test_prompt.py
ds-sebastianchwilczynski: okay, it makes sense

packages/ragbits-core/src/ragbits/core/prompt/prompt.py
sbetlewski: simple and works, perfect 💪🏼

ds-sebastianchwilczynski (approved):

PR #724, opened 2025-07-16
feat: Impersonation and application/pdf export to mime map for presentation type.

pocucan-ds:
## Creating a Feature to Impersonate Google Accounts 
- able to impersonate google accounts via google-service-accounts.
-  moved the mapping from powerpoints to pdfs as it requires paid services to process slides from powerpoints. 
image


pocucan-ds:
@maxpill  finished my PR while i was temporarily suspended

PR #727, opened 2025-07-18
feat: Autogenerate TS types

jakubduda-dsai:


packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -132,32 +119,28 @@ async def feedback(request: FeedbackRequest) -> JSONResponse:
...
+                ),
+                customization=self.chat_interface.ui_customization,
+                user_settings=self.chat_interface.user_settings,
+                debug_mode=self.debug_mode,
+                conversation_history=self.chat_interface.conversation_history,
+            )
+
+            return JSONResponse(content=config_response.model_dump())
+
+    async def _handle_chat_message(self, request: ChatRequest) -> StreamingResponse:  # noqa: PLR0915
```
dazy-ds: nit: You might want to ask Mateusz about the need for backward compatibility for this `ChatMessageRequest` type. Maybe we should keep it with deprecation notice?

packages/ragbits-chat/src/ragbits/chat/client/conversation.py
```diff
@@ -10,7 +10,7 @@
     ChatResponse,
     ChatResponseType,
     Message,
-    MessageRole,
```
dazy-ds: nit: Same as with the `ChatMessageRequest` type. Let's ask Mateusz about compatibility.

packages/ragbits-chat/src/ragbits/chat/interface/types.py
```diff
@@ -145,10 +157,56 @@ def as_image(self) -> Image | None:
...
+    status: str = Field(..., description="Status of the feedback submission")
+
+
+class FeedbackRequest(BaseModel):
+    """
+    Request body for feedback submission
+    """
+
+    message_id: str = Field(..., description="ID of the message receiving feedback")
+    feedback: Literal["like", "dislike"] = Field(..., description="Type of feedback (like or dislike)")
```
dazy-ds: Shouldn't we use `FeedbackType` instead of `Literal[...]`?

typescript/ui/__tests__/integration/intergation.test.tsx
```diff
@@ -56,6 +55,16 @@ const historyStore = createStore(createHistoryStore);
...
 
+const chatResponseTypes: ChatResponseType[] = [
+  "text",
+  "reference",
+  "state_update",
+  "message_id",
+  "conversation_id",
+  "live_update",
+  "followup_messages",
+] as const;
```
dazy-ds: This is one of the things I considered when we decided to move away from enums. The problem with this array is that it doesn't enforce the constraints that we want at all. We can omit all types from this array and still don't get feedback from the compiler...

Let's maybe change enum generation to something like:
```
export const ChatResponseType = {
  Text: 'text',
  Reference: 'reference',
  StateUpdate: 'state_update',
  MessageId: 'message_id',
  ConversationId: 'conversation_id',
  LiveUpdate: 'live_update',
  FollowupMessages: 'followup_messages',
  Image: 'image',
} as const;

export type ChatResponseType = typeof ChatResponseType[keyof typeof ChatResponseType];
```
which almost reimplements the `enum` keyword, but now we are explicit in defining the object (so it doesn't magically appear in our compiled code). Or alternatively we can still go back to enums.

packages/ragbits-chat/src/ragbits/chat/interface/types.py
jakubduda-dsai: I think so, my bad

Issue #729, opened 2025-07-21
feat: add images to UI


mhordynski:
### Feature description

..

### Motivation

..

PR #731, opened 2025-07-21
feat: #729 - images support in chat message, image gallery

jakubduda-dsai:


typescript/ui/src/core/components/ChatMessage/MessageActions.tsx
```diff
@@ -0,0 +1,76 @@
...
+      clearTimeout(copyIconTimerRef.current);
+    }
+    copyIconTimerRef.current = setTimeout(() => {
+      setCopyIcon("heroicons:clipboard");
+    }, 2000);
+  };
+
+  return (
+    <div className="flex items-center gap-2">
+      {!isLoading && (
```
dazy-ds: nit: In my opinion the component shouldn't be aware of `isLoading` as it doesn't use it anywhere but here. I think this condition should be in the parent `{isLoading && <MessageActions .../>}`. WDYT?

typescript/ui/src/types/history.ts
```diff
@@ -21,6 +22,7 @@ export interface ChatMessage {
   references?: Reference[];
   liveUpdates?: Map<string, LiveUpdate["content"]>;
   extensions?: Record<string, unknown>;
+  images?: Map<string, Image["url"]>;
```
dazy-ds: Along with the history I changed all maps to be records (it's easier to serialize and we don't really use Maps properties). What do you think about usage of `Record<string, Image['url']>` here?

dazy-ds (approved):

typescript/ui/src/core/components/ChatMessage/MessageActions.tsx
jakubduda-dsai: sure!

typescript/ui/src/types/history.ts
jakubduda-dsai: oki, no problem at all

PR #732, opened 2025-07-21
feat(ui): conversation history in UI

dazy-ds:


jakubduda-dsai:
A few things based on how UI looks like/works
- bottom left corner is not rounded
- I think I am not a fan of that when I close app and reopen it automatically reads last conversation. I'd prefer empty conv. WDYT?
- WDYT to make open exact conv based on routing like domain.com/conv/:id or so?
- The clear button is now missunderstanding, for a while I was looking for a way to start new conv. WDYT about moving it to sidebar and creating "New chat" button

Also I'd see sidebar with conversations like debug sidebar, but  not hiding it completely just collapsing to minimal version with create new chat button or expanded with conv history + new chat button

typescript/ui/src/core/components/ChatHistory.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export default function ChatHistory() {
+  const { selectConversation, deleteConversation } = useHistoryActions();
+  const conversations = useHistoryStore((s) => s.conversations);
+  const currentConversation = useHistoryStore((s) => s.currentConversation);
+
+  return (
+    <div className="max-w-64 flex-grow overflow-auto border-b-small border-l-small border-divider p-4">
+      <p className="truncate text-small font-semibold leading-5 text-foreground">
+        Previous conversations
```
jakubduda-dsai: What about just changing it to simply `Conversations`? WDYT?

dazy-ds:
> A few things based on how UI looks like/works
> 
> * bottom left corner is not rounded
> * I think I am not a fan of that when I close app and reopen it automatically reads last conversation. I'd prefer empty conv. WDYT?
> * WDYT to make open exact conv based on routing like domain.com/conv/:id or so?
> * The clear button is now missunderstanding, for a while I was looking for a way to start new conv. WDYT about moving it to sidebar and creating "New chat" button
> 
> Also I'd see sidebar with conversations like debug sidebar, but not hiding it completely just collapsing to minimal version with create new chat button or expanded with conv history + new chat button

Ad. 1. Good catch, will fix.
Ad. 2. We can change that.
Ad. 3. I don't have any strong opinions. For now it seems like unnecessary complexity. We can consider adding this with addition of router (that will likely come during auth integration). What do you think about that?
Ad. 4. For me it seemed fine as it's functionality didn't change at all, but moving is a good idea.

EDIT:
I intentionally did the sidebar this way to always preserve the history visible as in my opinion it better fits into the layout. I am fine with either approach, maybe Mateusz has some strong opinions in that matter.

EDIT 2:
I addressed 1., 2. and 4. Let's discuss 3. and your last remark via chat.

@jakubduda-dsai

typescript/ui/src/types/history.ts
```diff
@@ -19,20 +17,24 @@ export interface ChatMessage {
...
   content: string;
   references?: Reference[];
-  liveUpdates?: Map<string, LiveUpdate["content"]>;
+  liveUpdates?: Record<string, LiveUpdate["content"]>;
   extensions?: Record<string, unknown>;
 }
 
-export interface HistoryStore {
-  history: Map<string, ChatMessage>;
+export interface ConversationHistory {
```
jakubduda-dsai: WDYT about naming it Conversation? I think about this because it is one conv with all data

dazy-ds: Fixed

typescript/ui/src/core/components/ChatHistory.tsx
dazy-ds: Changed

PR #733, opened 2025-07-21
feat: Google drive impersonator

maxpill:


packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -62,12 +86,31 @@ class GoogleDriveSource(Source):
...
+    impersonate: ClassVar[bool | None] = None
+    impersonate_target_email: ClassVar[str | None] = None
 
     @classmethod
     def set_credentials_file_path(cls, path: str) -> None:
         """Set the path to the service account credentials file."""
         cls._credentials_file_path = path
 
+    @classmethod
+    def set_impersonation_target(cls, target_mail: str) -> None:
```
mhordynski: I think it should be instance level property, not on a class level.

Why? Because user may want to impersonate as different account to reach specific file

docs/how-to/sources/google-drive.md
```diff
@@ -187,6 +187,43 @@ async def process_drive_documents():
...
+
+### Step 2: Impersonate a User in Your Code
+
+After configuring domain-wide delegation, you can specify a target user to impersonate when using the `GoogleDriveSource` in your code.
+
+```python
+from ragbits.core.sources.google_drive import GoogleDriveSource
+
+target_email = "johnDoe@yourdomain.com"
+credentials_file = "service-account-key.json"
```
ds-sebastianchwilczynski, line 212: I'd put link to the part about creating this credential_file here. 

This how-to is very long and I believe someone may forgot what this credential file is all about long ago. Or if they come directly to this place, they're cooked

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -62,12 +86,31 @@ class GoogleDriveSource(Source):
...
+        Sets the email address to impersonate when accessing Google Drive resources.
+
+        Args:
+            target_mail (str): The email address to impersonate.
+
+        Raises:
+            ValueError: If the provided email address is invalid (empty or missing '@').
+        """
+        # check if email is a valid email.
+        if not target_mail or "@" not in target_mail:
```
ds-sebastianchwilczynski: can we use a regex for this @.@?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -62,12 +86,31 @@ class GoogleDriveSource(Source):
...
+            raise ValueError("Invalid email address provided for impersonation.")
+        cls.impersonate = True
```
ds-sebastianchwilczynski: Do we need this variable at all, couldn't we just see if impersonate_target_email is None below?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -162,7 +218,11 @@ def verify_drive_api_enabled(cls) -> None:
 
     @traceable
     @requires_dependencies(["googleapiclient"], "google_drive")
-    async def fetch(self) -> Path:
+    async def fetch(
+        self,
+        *,
+        export_format: "GoogleDriveExportFormat | None" = None,
```
ds-sebastianchwilczynski: Why is this in quotes?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -186,7 +249,8 @@ async def fetch(self) -> Path:
         file_local_dir = local_dir / self.file_id
         file_local_dir.mkdir(parents=True, exist_ok=True)
 
-        export_mime_type, file_extension = self._determine_file_extension()
+        override_mime = export_format.value if export_format else None
```
ds-sebastianchwilczynski: Why do we extract value from export_format? Can't we pass it directly to _determine_file_extension, and expect this Enum type? 

I don't get it because in _determine_file_extension we make it Enum again

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -62,12 +86,31 @@ class GoogleDriveSource(Source):
 
     _google_drive_client: ClassVar["GoogleAPIResource | None"] = None
     _credentials_file_path: ClassVar[str | None] = None
+    impersonate: ClassVar[bool | None] = None
```
ds-sebastianchwilczynski: So if we want to make this instance property. Impersonate (imo not needed) and impersonate_target_email shouldn't be class variables here anymore as then this will be non thread safe.

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -186,7 +249,8 @@ async def fetch(self) -> Path:
         file_local_dir = local_dir / self.file_id
         file_local_dir.mkdir(parents=True, exist_ok=True)
 
-        export_mime_type, file_extension = self._determine_file_extension()
+        override_mime = export_format.value if export_format else None
+        export_mime_type, file_extension = self._determine_file_extension(override_mime=override_mime)
         local_file_name = f"{self.file_name}{file_extension}"
         path = file_local_dir / local_file_name
```
ds-sebastianchwilczynski, line 277: If my file has name `file.pdf` will it be saved as `file.pdf.pdf`?

packages/ragbits-core/src/ragbits/core/sources/google_drive.py
```diff
@@ -496,22 +560,36 @@ async def from_uri(cls, path: str) -> Iterable[Self]:
...
         Determine the appropriate file extension and export MIME type for the file.
 
         Returns:
             A tuple of (export_mime_type, file_extension)
         """
+        if override_mime is not None:
+            export_mime_type = override_mime
+            try:
+                export_format = GoogleDriveExportFormat(override_mime)
+                file_extension = _EXPORT_EXTENSION_MAP.get(export_format, ".bin")
```
ds-sebastianchwilczynski: Probably not for this Pr but why do we assume .bin is a default extension?

ds-sebastianchwilczynski (changes requested):
Please modify this impersonate_target_email to make it an instance property.

For other stuff I'm open to discussion

PR #744, opened 2025-07-23
refactor(ui): redesign history UI, allow history configuration in config

dazy-ds:


examples/chat/chat.py
```diff
@@ -86,6 +86,8 @@ class MyChat(ChatInterface):
         ),
     )
 
+    client_side_history = True
```
jakubduda-dsai: I'd suggest to rename it to conversations_history or so and treat is as boolean for now and in future it will be changed prbbly to Object when there will be possibility to choose between backend_side and client_side history. In future I'd see setting config for history in one place

jakubduda-dsai (approved):

examples/chat/chat.py
dazy-ds: Sure, changed.

PR #748, opened 2025-07-25
fix: missing model names in model cost litellm dict

ds-sebastianchwilczynski:
When you initialise your llm like this

`model = LiteLLM(model_name="anthropic/claude-3-7-sonnet-20250219", default_options=options)`

you'll get exception on missing keys in `model_cost` litellm's dictionary


pocucan-ds (approved):
easy approval.

PR #752, opened 2025-07-25
feat: add reasoning

ds-sebastianchwilczynski:


packages/ragbits-core/src/ragbits/core/llms/litellm.py
```diff
@@ -45,6 +47,8 @@ class LiteLLMOptions(LLMOptions):
     mock_response: str | None | NotGiven = NOT_GIVEN
     tpm: int | None | NotGiven = NOT_GIVEN
     rpm: int | None | NotGiven = NOT_GIVEN
+    reasoning_effort: str | None | NotGiven = NOT_GIVEN
```
ghost: I think some documentation for these would be nice I have no idea what to pass there

packages/ragbits-core/CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- Add support for Reasoning models (#752)
```
ghost, line 5: maybe let's add an example / modify one of existing ones to showcase how to use reasoning model?

packages/ragbits-core/src/ragbits/core/llms/litellm.py
ghost: I tried to use it like this:
```
    llm = LiteLLM(model_name="o3-mini", use_structured_output=True)
    response = await llm.generate_with_metadata("What is the meaning of life?", options=LiteLLMOptions(reasoning_effort="medium"))
    print(response)
```
but my response has resoning=None

ds-sebastianchwilczynski: Above there is a link to litellm documentation on reasoning but will add something additional

ds-sebastianchwilczynski: The thing is [OpenAI will not return reasoning content for you](https://platform.openai.com/docs/guides/reasoning?api-mode=responses)
image

You must try anthropic models

ghost: Ah sorry, by accident I copy pasted the link to the input params and didn't see reasoning there :sweat_smile: 
The thing with OpenAI is interesting, I wonder how we can let users know about this. Maybe in case of OpenAI models we should put some information that reasoning won't be provided? Because it's look like a bug when you specify reasoning effort and get None as output. Or maybe some kind of warning in logs?

ds-sebastianchwilczynski: I put info about this in our documentation.

PR #754, opened 2025-07-29
feat(ui): client routing

dazy-ds:


typescript/ui/src/plugins/ChatHistoryPlugin/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+export const ChatHistoryPluginName = "ChatHistoryPlugin";
+export const ChatHistoryPlugin = createPlugin({
+  name: ChatHistoryPluginName,
+  components: {
+    ChatHistory: lazy(() => import("./components/ChatHistory")),
+  },
+  routes: [
+    {
+      target: `/`,
+      path: `/history/:conversationId`,
```
jakubduda-dsai: WDYT about naming it /conversation/:conversationId instead of history?

typescript/ui/src/plugins/ChatHistoryPlugin/routes/HistoryGuard.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { getConversationRoute } from "../utils";
+
+export default function HistoryGuard({ children }: PropsWithChildren) {
+  const conversations = useHistoryStore(
+    useShallow((s) => Object.keys(s.conversations)),
+  );
+  const { selectConversation } = useHistoryActions();
+  const params = useParams();
+  const conversationId = params.conversationId;
+  const isValidURL =
```
jakubduda-dsai: I think that name of this variable is kind of missleading. It doesnt check if url is valid but conversationId

typescript/ui/src/plugins/ChatHistoryPlugin/index.tsx
dazy-ds: Either way is fine, changed.

typescript/ui/src/plugins/ChatHistoryPlugin/routes/HistoryGuard.tsx
dazy-ds: Changed

PR #761, opened 2025-08-01
feat(auth): Backend authentication into the chat

GlockPL:
Adds Authentication Backend and integrates it with the Ragbits API.


jakubduda-dsai:
Hi, I ran BE as usually for other examples with pattern like `ragbits api run examples.chat.authenticated_chat:MyAuthenticatedChat` and for authenticated chat example this not work, @dazy-ds told me that it needs to be run with command `python examples/chat/authenticated_chat.py`. Can we do it like other examples or is it not possible?

GlockPL:
Ok, I'll try integrating this with CLI, so that it can be run properly from cmd

GlockPL:
Ok, now it works as intended, to test you can try
`uv run ragbits api run examples.chat.authenticated_chat:MyAuthenticatedChat --auth examples.chat.authenticated_chat:get_auth_backend`

examples/chat/authenticated_chat.py
```diff
@@ -0,0 +1,336 @@
...
+    feedback: str = Field(description="Please provide more details", min_length=1)
+
+
+class UserSettingsFormExample(BaseModel):
+    """A simple example implementation of the chat form that demonstrates how to use Pydantic for form definition."""
+
+    model_config = ConfigDict(title="Chat Form", json_schema_serialization_defaults_required=True)
+
+    language: Literal["English", "Polish"] = Field(description="Please select the language", default="English")
+
```
jakubduda-dsai: One suggestion, this seems like copy from other chat examples. Maybe now is a good time to create some defaults.py file or templates.py or so, and just import this basic stuff instead of duplicate?

packages/ragbits-chat/src/ragbits/chat/interface/types.py
```diff
@@ -202,6 +203,13 @@ class FeedbackConfig(BaseModel):
     dislike: FeedbackItem = Field(..., description="Dislike feedback configuration")
 
 
+class AuthenticationConfig(BaseModel):
+    """Configuration for authentication."""
+
+    enabled: bool = Field(default=False, description="Enable/disable authentication")
+    login_forms: list[str] = Field(default=[], description="List of authentication forms")
```
jakubduda-dsai: should it be `list[str]` or list of exact strings like `JWTAuth, GoogleAuth` etc

maybe we should provide additional type lets say `AuthTypes` and name this fields `auth_types: list[AuthType]`

also naming `forms` is used in project to send schemas for displaying forms, I'd suggest naming this login_types

examples/chat/README_authenticated.md
```diff
@@ -0,0 +1,196 @@
...
+- `README_authenticated.md` - This documentation
+
+## Quick Start
+
+### 1. Start the Server
+
+#### Full Authentication Support (Recommended)
+```bash
+# From the examples/chat directory - includes login/logout endpoints
+uv run python authenticated_chat.py
```
jakubduda-dsai, line 29: is this command still valid?

examples/chat/README_authenticated.md
```diff
@@ -0,0 +1,196 @@
...
+# From project root
+python examples/chat/authenticated_chat.py
+
+# Via CLI (limited auth support)
+uv run ragbits api run examples.chat.authenticated_chat:MyAuthenticatedChat
+```
+
+The server will start at `http://127.0.0.1:8000` with a web interface.
+
+**Note**: The CLI version has limited authentication support because it doesn't create the authentication endpoints (`/api/auth/login`, `/api/auth/logout`). For full authentication features, use the direct Python execution.
```
jakubduda-dsai: i think this not should be updated afaik now there is possibility to add --auth and via cli also everything would work. Am I right?

packages/ragbits-chat/src/ragbits/chat/auth/base.py
```diff
@@ -0,0 +1,86 @@
...
+from .models import JWTToken, OAuth2Credentials, User, UserCredentials
+
+
+class AuthBackendOptions(Options):
+    """Options for authentication backends."""
+
+    pass
+
+
+class AuthenticationResult(BaseModel):
```
jakubduda-dsai: If this is what api return maybe it would be good to stick with current naming, like Response/Request etc

examples/chat/README_authenticated.md
GlockPL: Yes, it can be run in both ways. I can remove this option if needed.

examples/chat/README_authenticated.md
GlockPL: Yes, sorry, forgot about that

packages/ragbits-chat/src/ragbits/chat/interface/types.py
GlockPL: Ok, this can be easly added

packages/ragbits-chat/src/ragbits/chat/auth/base.py
GlockPL: Changed

examples/chat/README_authenticated.md
GlockPL: Fixed

packages/ragbits-chat/src/ragbits/chat/interface/types.py
GlockPL: Changed

examples/chat/authenticated_chat.py
GlockPL: Here I'm not so sure, my understanding is that every example is self contained, that's why there are repetitions

jakubduda-dsai: @mhordynski WDYT?

mhordynski: let's remove forms from this example

GlockPL: Removed

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
```diff
@@ -0,0 +1,184 @@
...
+        Args:
+            users: List of user dicts with 'username', 'password', and optional fields
+            jwt_secret: Secret key for JWT jwt_token signing (generates random if not provided)
+            default_options: Default options for the component
+        """
+        if default_options is None:
+            default_options = AuthOptions()
+        super().__init__(default_options)
+        self.users = {}
+        self.jwt_secret = jwt_secret or secrets.token_urlsafe(32)
```
maxpill: all sessions invalidated on every server restart

mhordynski: use here `ragbits.core.utils.secrets` logic

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
```diff
@@ -0,0 +1,184 @@
...
+
+        Args:
+            token: The JWT token to revoke
+
+        Returns:
+            True if successfully revoked, False if token is invalid
+        """
+        try:
+            # Validate the token first to ensure it's a valid JWT
+            jwt.decode(token, self.jwt_secret, algorithms=[self.jwt_algorithm])
```
mhordynski: this should raise exception in case of ListBackend - rn it wouldn't work and can lead to dangerous situations in case of more than 1 instance of the app

examples/chat/authenticated_chat.py
GlockPL: removed

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
GlockPL: Implemented

GlockPL: Implemented with get_secret_key()

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
GlockPL: Implemented

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
```diff
@@ -0,0 +1,173 @@
...
+        """
+        Revoke a JWT token.
+
+        Args:
+            token: The JWT token to revoke
+
+        Raises:
+            NotImplementedError: This method is not implemented
+        """
+        raise NotImplementedError("revoke_token method is not implemented")
```
mhordynski:
```suggestion
        raise NotImplementedError("ListAuthenticationBackend is designed to run in development / testing scenarios. Revoking tokens is not implemented in this backend, if you need to revoke tokens please consider using different backend or implementing your own.")
```

PR #763, opened 2025-08-04
feat(ui): initial auth plugin

dazy-ds:


typescript/@ragbits/api-client/src/types.ts
```diff
@@ -189,8 +189,48 @@ export interface ConfigResponse {
     }
     debug_mode: boolean
     conversation_history: boolean
+    authentication: {
+        enabled: boolean
+        type: string | null
```
jakubduda-dsai: Maybe to early but, WDYT about type: string[] which would suggest empty array as nothing obiously and in array user could pass few auth types, like google/user+pass/github etc? I know that for now we would have only one type but for future it would be easier to provide compatibility and extend it

typescript/@ragbits/api-client/src/types.ts
```diff
@@ -189,8 +189,48 @@ export interface ConfigResponse {
...
+
+export interface JWTToken {
+    access_token: string
+    token_type: string
+    expires_in: number
+    refresh_token: string | null
+    user: User
+}
+
+export interface LoginRequest {
```
jakubduda-dsai: If we decide to think about future (earlier comment) maybe it wouold be good idea also to have sth like
type LoginRequest = JWTLoginRequest | GoogleLoginRequest  etc

typescript/ui/src/plugins/AuthPlugin/utils/useAuthenticatedCall.tsx
```diff
@@ -0,0 +1,21 @@
+import { useRagbitsCall } from "@ragbits/api-client-react";
+import { useAuthStore } from "../contexts/AuthStoreContext/useAuthStore";
+import { produce } from "immer";
+
+const useAuthenticatedCall: typeof useRagbitsCall = (
```
jakubduda-dsai, line 5: Hmmm... Have you thought about new method/hook in npm ragbits packages like authenticatedCall/useAuthenticatedCall for auth purposes? So if people would build their own custom frontend apps they would not need to write a function like this? WDYT?

jakubduda-dsai:
One thing to fix actually

I ran dev environment (backend example authenticated_chat.py) and from `/` I was redirected to `/conversation/temp-:uuid`
I am able to see message about secured chat, dont get response when I send message (expected)
BUT
1. I am not redirected to /login or I don't see button to go to login page, so I can't login
2. When I start new conversation and type message, request is send to backend and the conversation id is returned and conv is created in conv list. Not sure if this is expected

jakubduda-dsai (changes requested):
Previous comment

typescript/@ragbits/api-client/src/types.ts
dazy-ds: This type comes explicitly from the backend. In terms of extensibility backend can always extend the type to include some kind of `authType` field which can tell us what to expect without breaking this type in older clients (as in case of missing `authType` we just treat the field the old way).

typescript/@ragbits/api-client/src/types.ts
dazy-ds: Same response as above C:

typescript/ui/src/plugins/AuthPlugin/utils/useAuthenticatedCall.tsx
dazy-ds: I thought about this but I decided to keep it in the `ui` package, as the libraries should be as generic as possible (hence the addition of `headers` argument). Also the authentication methods may vary in custom frontend implementation and our hook uses our `authStore`. I would keep it as is right now, we can always move it if needed.

dazy-ds:
> Previous comment

For auth to work, one should start the server with the following command `python examples/chat/authenticated_chat.py` C: I think that should be changed to align with other examples but it's part of the BE review.

jakubduda-dsai:
> > Previous comment
> 
> For auth to work, one should start the server with the following command `python examples/chat/authenticated_chat.py` C: I think that should be changed to align with other examples but it's part of the BE review.

cc: @GlockPL

typescript/@ragbits/api-client/src/types.ts
jakubduda-dsai: @GlockPL @mhordynski WDYT?

typescript/@ragbits/api-client/src/types.ts
jakubduda-dsai: @GlockPL @mhordynski WDYT?

typescript/ui/src/plugins/AuthPlugin/utils/useAuthenticatedCall.tsx
jakubduda-dsai: Yea, I thought about it again after I added the comment and I decided as you haha 😜 Let's keep it this way

typescript/@ragbits/api-client/src/types.ts
GlockPL: I've created the AuthBackend class in a way that would potentially allow for a future versions with different Login approach

PR #769, opened 2025-08-08
Tests added for parallel batches execution

vladimir-kivi-ds:


packages/ragbits-evaluate/src/ragbits/evaluate/evaluator.py
```diff
@@ -71,6 +71,7 @@ def __init__(
         num_retries: int = 3,
         backoff_multiplier: int = 1,
         backoff_max: int = 60,
+        parallel_batches: bool = False,
```
mhordynski: hmm, the name `parallel_batches` makes me think it should be `int` type.

Maybe `parallelize_batches`?  

On the other hand, maybe it makes sense to have it as `int` to control how many tasks we sen to  gather at once.

mhordynski: ah okay just realized that number of tasks is implicitly controlled by batch_size, so omit that last suggestion

vladimir-kivi-ds: Alright, makes sense

PR #776, opened 2025-08-11
feat: Agents in CLI

GlockPL:
Ability to chat with agent in CLI by passing path to the Agent instance in terminal


packages/ragbits-agents/src/ragbits/agents/cli.py
```diff
@@ -0,0 +1,472 @@
...
+    if ":" not in agent_path:
+        raise ValueError(
+            f"Invalid agent path format: {agent_path}. "
+            f"Expected 'module.path:agent_variable' or 'path/to/file.py:agent_variable'"
+        )
+
+    module_path, agent_name = agent_path.split(":", 1)
+
+    # Check if it's a file path
+    if module_path.endswith(".py") and (Path(module_path).exists() or "/" in module_path or "\\" in module_path):
```
pocucan-ds, line 40: Does the agent know by default what OS system it's using? is that why we are checking these multiple statements? 


I wonder what would happen if we ran this agent on windows OS. 

Given that a lot of people use Windows, we probably should check if agents are able to run cli commands on windows? 

What do you think.

GlockPL: I would have to get my hands on windows machine, I could try my private one.

packages/ragbits-agents/src/ragbits/agents/cli.py
```diff
@@ -0,0 +1,472 @@
...
+                    result = await self.agent.run(message)
+
+                    # Clear the log and rewrite without the thinking message
+                    last_lines = list(chat_log.lines)
+                    if last_lines and "*thinking...*" in str(last_lines[-1]):
+                        # Remove the thinking message by clearing and rewriting all but the last line
+                        chat_log.clear()
+                        for line in last_lines[:-1]:
+                            chat_log.write(str(line))
+
```
maxpill: The current approach adds a "thinking..." message to the chat log, then clears the entire log and rewrites all lines except the thinking message when done. 

We're doing a full log clear/rewrite just to remove one line. The string matching logic is also error prone, checking if "thinking..." exists in the last line. We have duplicate cleanup code in both success and error paths.

Thinking about our future development - **this approach will make it hard to add streaming thoughts or progressive updates without major refactoring**.

I suggest we use a dedicated status component (i.e. Static from textual lib) for thinking animation instead of polluting chat history. Add a component that shows i.e. animated dots using an asyncio (or set_interval from textual) task cycling through "thinking", "thinking.", "thinking..", "thinking..." or anything else. When done, cancel the animation and add the response to chat.

With such approach chat log maintains conversation history, status widget handles temporary feedback. No log manipulation needed, chat history stays clean - no need for history manipulation, error handling is simple (just cancel the task), and it's easy to extend for streaming thoughts by updating the status message.

maxpill: @mhordynski

packages/ragbits-agents/src/ragbits/agents/cli.py
```diff
@@ -0,0 +1,472 @@
...
+                chat_log = self.query_one("#chat_log", Log)
+                tools_log = self.query_one("#tools_log", Log)
+
+                try:
+                    # Show a temporary thinking message
+                    chat_log.write_line("🤖 Agent: *thinking...*")
+                    self.refresh()  # Force UI refresh to show thinking message
+
+                    # Run the agent
+                    result = await self.agent.run(message)
```
mhordynski: why not run_streaming?

mhordynski: and remove thiking... message. We can add some spinner or smth like that but it's "nice to have"

packages/ragbits-agents/src/ragbits/agents/cli.py
mhordynski: I've replied here: https://github.com/deepsense-ai/ragbits/pull/776/files/46b7776680b8b28a5e25b0cea958ff7b8de55479#r2269903791

packages/ragbits-agents/src/ragbits/agents/cli.py
GlockPL: I've added a spinner and moved to run_streaming

packages/ragbits-agents/src/ragbits/agents/cli.py
GlockPL: Thinking was removed and spinner was added

packages/ragbits-agents/src/ragbits/agents/cli.py
GlockPL: Ok, tested it on windows 11, works fine

packages/ragbits-agents/src/ragbits/agents/cli.py
GlockPL: Switched to run_streaming

packages/ragbits-agents/src/ragbits/agents/cli.py
pocucan-ds: approved.

Issue #734, opened 2025-07-22
docs: create chat interface tutorial


mhordynski:
We should create a new tutorial in ragbits on how-to create user interfaces with `ragbits-chat`

It should go through following features:

* how to start user interface
* how to return text responses
* how to return references, images, follow_ups
* how to return live_updates
* how to store state
* how to configure user forms
* how to setup custom icon, header

PR #778, opened 2025-08-12
feat: #734 - how to for chat

jakubduda-dsai:


examples/chat/tutorial.py
```diff
@@ -0,0 +1,233 @@
...
+        model_name: The model name to use
+
+    Returns:
+        A callable async web search function that returns content and references
+    """
+    from ragbits.agents.tools.openai import OpenAIResponsesLLM
+
+    async def search_web_with_references(query: str) -> dict:
+        # Create the OpenAI tool instance
+        openai_tool = OpenAIResponsesLLM(model_name, {"type": "web_search_preview"})
```
mhordynski: @mhordynski

typescript/@ragbits/api-client/src/index.ts
```diff
@@ -229,6 +250,70 @@ export class RagbitsClient {
...
+                mimeType,
+            })
+        }
+
+        // Store the chunk
+        const imageInfo = this.imageChunks.get(id)!
+        imageInfo.chunks.set(chunkIndex, chunkData)
+
+        // Check if all chunks are received
+        if (imageInfo.chunks.size === totalChunks) {
```
dazy-ds: I would reverse this condition and just use early return.

typescript/@ragbits/api-client/src/index.ts
```diff
@@ -14,6 +15,14 @@ import type {
...
 export class RagbitsClient {
     private readonly baseUrl: string
+    private imageChunks: Map<
+        string,
+        {
+            chunks: Map<number, string>
+            totalChunks: number
+            mimeType: string
+        }
+    > = new Map()
```
dazy-ds: I am not sure about the naming. Since we pass the `mime_type` in the chunk, we have a mechanism to handle all kinds of chunked data, not only images. Let's change this name to something like `chunk_queue` (or something in that manner) so it is more generic.

typescript/@ragbits/api-client/src/index.ts
```diff
@@ -229,6 +250,70 @@ export class RagbitsClient {
...
+                    type: ChatResponseType.Image,
+                    content: {
+                        id: id,
+                        url: `${imageInfo.mimeType},${completeBase64}`,
+                    },
+                }
+                // Send the complete image
+                await callbacks.onMessage(completeImageResponse as T)
+            }
+        }
```
dazy-ds, line 319: One more thing I would add is removal of the chunks from the `imageChunks` map when we construct the whole message. There is no need to hold it in two places at the same time (here and in the history)

typescript/@ragbits/api-client/src/index.ts
```diff
@@ -229,6 +250,70 @@ export class RagbitsClient {
...
+        data: T,
+        callbacks: StreamCallbacks<T>
+    ): Promise<void> {
+        const response = data as ChunkedChatResponse
+        const content = response.content
+        const contentType = content.content_type
+        const id = content.id
+        const chunkIndex = content.chunk_index
+        const totalChunks = content.total_chunks
+        const mimeType = content.mime_type
```
dazy-ds: What do you think about using destructuring assignment instead?

typescript/@ragbits/api-client/src/index.ts
jakubduda-dsai: yeah, I changed other fields but forgot about this one :/

typescript/@ragbits/api-client/src/index.ts
jakubduda-dsai: sure

dazy-ds (approved):
FE part looks fine

mhordynski (approved):

mhordynski (approved):

PR #819, opened 2025-09-12
feat: support wrapping downstream agents as tools

akotyla:


packages/ragbits-agents/src/ragbits/agents/tool.py
```diff
@@ -96,5 +100,61 @@ def to_pydantic_ai(self) -> "PydanticAITool":
             description=self.description,
         )
 
+    @classmethod
+    def from_agent(cls, agent: "Agent") -> "Tool":
```
mhordynski: let's allow to overwrite default agent name and description, it may happen that i want to optimize for a specific agent how it is presented as a tool

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -219,6 +219,8 @@ class Agent(
     def __init__(
```
mhordynski, line 219: I think it would be handy to have also agent->tool method on this side too, i mean Agent.to_tool(...)

prob it would require local import / if TYPE_CHECKING to mitigate circular imports

PR #820, opened 2025-09-15
feat: class-based agents

dazy-ds:


mhordynski:
I think it would be best for now separate this PR to two:

1. With explicit input_type passed to the class as this is more straightforward and could be merged already
2. Second one with input_ logic - and there we can spend more time on thinking if we can solve it better

dazy-ds:
> I think it would be best for now separate this PR to two:
> 
> 1. With explicit input_type passed to the class as this is more straightforward and could be merged already
> 2. Second one with input_ logic - and there we can spend more time on thinking if we can solve it better

As requested and agreed via other channels, this PR contains the version with `prompt_config` decorator.
It is ready to be reviewed C: @mhordynski

mhordynski (approved):

PR #821, opened 2025-09-15
feat: introduce post processors

mackurzawa:


packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+        If True, the processor can work with content during streaming
+        via process_streaming() method.
+
+        If False, the processor will only be called after streaming is complete
+        with the full result via process() method.
+        """
+        return False
+
+    async def process(  # noqa: PLR6301
```
akotyla: Instead of suppressing `PLR6301`, could we mark this method as `@abstractmethod` and leave it without this base implementation?

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
```diff
@@ -0,0 +1,70 @@
...
+            result: The complete AgentResult from the agent or previous post-processor.
+            agent: The Agent instance that generated the result. Can be used to re-run
+                  the agent with modified input if needed.
+
+        Returns:
+            Modified AgentResult to pass to the next processor or return as final result.
+        """
+        return result
+
+    async def process_streaming(  # noqa: PLR6301
```
akotyla: Same here :D can we mark this as `@abstractmethod` with no base implementation?

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
```diff
@@ -0,0 +1,70 @@
...
+from ragbits.core.llms.base import LLMOptions, ToolCall, Usage
+from ragbits.core.prompt.base import BasePrompt
+
+if TYPE_CHECKING:
+    from ragbits.agents._main import Agent, AgentResult
+
+
+LLMOpts = TypeVar("LLMOpts", bound=LLMOptions)
+PromptIn = TypeVar("PromptIn", bound=BaseModel | None)
+PromptOut = TypeVar("PromptOut")
```
akotyla: we can import it from here: https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-core/src/ragbits/core/prompt/base.py#L9

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
```diff
@@ -0,0 +1,70 @@
...
+from ragbits.agents.tool import ToolCallResult
+from ragbits.core.llms.base import LLMOptions, ToolCall, Usage
+from ragbits.core.prompt.base import BasePrompt
+
+if TYPE_CHECKING:
+    from ragbits.agents._main import Agent, AgentResult
+
+
+LLMOpts = TypeVar("LLMOpts", bound=LLMOptions)
+PromptIn = TypeVar("PromptIn", bound=BaseModel | None)
```
akotyla: here also we can import it from here: https://github.com/deepsense-ai/ragbits/blob/main/packages/ragbits-core/src/ragbits/core/prompt/prompt.py#L23

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
```diff
@@ -0,0 +1,70 @@
...
+
+from ragbits.agents.tool import ToolCallResult
+from ragbits.core.llms.base import LLMOptions, ToolCall, Usage
+from ragbits.core.prompt.base import BasePrompt
+
+if TYPE_CHECKING:
+    from ragbits.agents._main import Agent, AgentResult
+
+
+LLMOpts = TypeVar("LLMOpts", bound=LLMOptions)
```
akotyla: maybe let's change the name to `LLMOptionsT` to be consistent with how we generally call these types of variables :)

akotyla:
I ran the example and everything worked perfectly! :muscle:  I think it would also be helpful to add a section about post-processors to our docs.

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -685,6 +731,72 @@ async def _extract_agent_skills(self) -> list["AgentSkill"]:
...
+                usage=usage,
+            )
+
+            current_result = agent_result
+            for processor in non_streaming_processors:
+                current_result = await processor.process(current_result, self)
+
+            yield current_result.usage
+            yield prompt_with_history
+            yield SimpleNamespace(
```
akotyla: Do we need the `SimpleNamespace` here? `current_result` already has content, metadata, and tool_calls, so I’m wondering why we don’t just yield it directly :thinking:

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -426,8 +463,17 @@ def run_streaming(
             AgentToolNotAvailableError: If the selected tool is not available.
             AgentInvalidPromptInputError: If the prompt/input combination is invalid.
             AgentMaxTurnsExceededError: If the maximum number of turns is exceeded.
+            AgentInvalidPostProcessorError: If the post-processor is invalid.
         """
-        generator = self._stream_internal(input, options, context, tool_choice)
+        if post_processors:
+            if not allow_non_streaming and any(not p.supports_streaming for p in post_processors):
+                raise AgentInvalidPostProcessorError(
```
mhordynski, line 523: i would like it to be enforced on type-checking rather than in runtime. To do so we can have two separate types and prepare overloads based on this argument

packages/ragbits-agents/src/ragbits/agents/_main.py
mackurzawa: We could probably yield it directly, but it would require adding another case to the `__anext__` in `AgentResultStreaming`, where the yielded element would be an `AgentResult`. I don't think this would make much of a difference, but if you insist, I can do it

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
mackurzawa: Done

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
mackurzawa: Done

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
mackurzawa: Done

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
mackurzawa: The post-processor can implement only one of the two methods because it is either a streaming or non-streaming processor. Adding an `@abstractmethod` to both would require implementing a method that wouldn't even be used

mackurzawa: Now that I think of it, maybe the best way to go is to create separate classes for streaming and non-streaming post-processors. That seems like the cleanest approach

packages/ragbits-agents/src/ragbits/agents/_main.py
akotyla: No, we can leave it as it is

packages/ragbits-agents/src/ragbits/agents/post_processors/base.py
mackurzawa: done

packages/ragbits-agents/src/ragbits/agents/_main.py
mackurzawa: Done

docs/how-to/agents/use_post_processors.md
```diff
@@ -0,0 +1,64 @@
+# How-To: Use Post-Processors with Ragbits Agents
+
+Ragbits Agents can be enhanced with post-processors to intercept, log, filter, and modify their outputs. This guide explains how to implement and use post-processors to customize agent responses.
+
+## Post-Processors Overview
+
+Ragbits provides two types of post-processors:
+
+- **StreamingPostProcessor**: Processes outputs as they are generated, suitable for real-time applications.
+- **NonStreamingPostProcessor**: Processes the final output after generation, ideal for batch processing.
```
mhordynski, line 10: just PostProcessor or SyncPostProcessor

docs/how-to/agents/use_post_processors.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Implementing a Post-Processor
+
+To create a post-processor, inherit from the appropriate base class (`StreamingPostProcessor` or `NonStreamingPostProcessor`) and implement the required method.
```
akotyla: we can add these classes to the API reference and link them here

examples/agents/post_processors.py
```diff
@@ -0,0 +1,94 @@
...
+            history=result.history,
+            usage=result.usage,
+        )
+
+
+async def main() -> None:
+    """
+    Run the example.
+    """
+    llm = LiteLLM("gpt-3.5-turbo")
```
mhordynski, line 80: let's stick to 4.1-mini as defaults, from now on

examples/agents/post_processors.py
```diff
@@ -0,0 +1,94 @@
...
+    agent = Agent(llm=llm, prompt="You are a helpful assistant.")
+    post_processors = [
+        CustomStreamingProcessor(forbidden_words=["python"]),
+        CustomNonStreamingProcessor(max_length=200),
+    ]
+    stream_result = agent.run_streaming("What is Python?", post_processors=post_processors, allow_non_streaming=True)
```
mhordynski, line 86: let's pass post_processors here directly as list literal, and make sure that mypy runs with no errors in this file

examples/agents/post_processors.py
```diff
@@ -0,0 +1,94 @@
...
+
+class CustomNonStreamingProcessor(NonStreamingPostProcessor):
+    """
+    Non-streaming post-processor that truncates the content.
+    """
+
+    def __init__(self, max_length: int = 200) -> None:
+        self.max_length = max_length
+
+    async def process(self, result: "AgentResult", agent: "Agent") -> "AgentResult":
```
mhordynski, line 56: "" is not required here for type hints

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -444,8 +509,17 @@ def run_streaming(
...
+                    reason="Non-streaming post-processors are not allowed when allow_non_streaming is False"
+                )
+            generator = self._stream_with_post_processing(input, options, context, tool_choice, post_processors)
```
mhordynski, line 519: Let's make this method a sort of decorator / wrapper, then it won't be required for it to understand internal concepts of agent .run_streaming - it will just iterate over already created generator and add post-processing logic. 

With this impl we could also move that logic out of agent to post_processor package

PR #823, opened 2025-09-16
feat: todo list for agent

jakubduda-dsai:


packages/ragbits-agents/src/ragbits/agents/tools/todo.py
```diff
@@ -0,0 +1,199 @@
...
+            return self.tasks[self.current_index]
+        return None
+
+    def advance_to_next(self):
+        """Move to next task."""
+        self.current_index += 1
+
+
+# Storage - just one todo list per agent run
+_current_todo: TodoList | None = None
```
mhordynski: this can't be a global, when 2 agents will be running concurrently it will break

jakubduda-dsai: changed that

packages/ragbits-agents/src/ragbits/agents/tools/todo.py
```diff
@@ -0,0 +1,219 @@
...
+from dataclasses import dataclass, field
+from enum import Enum
+from typing import Any, Literal, Callable
+
+
+class TaskStatus(str, Enum):
+    """Task status options."""
+    PENDING = "pending"
+    IN_PROGRESS = "in_progress"
+    COMPLETED = "completed"
```
rk-izak, line 25: Maybe additional statuses could be added to address some error recovery later on? i.e. if a given step fails, it can be either cancelled or attempted again with a slightly different approach?
```python
FAILED = "failed"
CANCELLED = "cancelled"
RETRYING = "retrying"
```

packages/ragbits-agents/src/ragbits/agents/tools/todo.py
```diff
@@ -0,0 +1,219 @@
...
+
+
+@dataclass
+class Task:
+    """Simple task representation."""
+    id: str
+    description: str
+    status: TaskStatus = TaskStatus.PENDING
+    order: int = 0
+    summary: str | None = None
```
rk-izak, line 39: Similarly, to ensure a given task is executed maybe we should add its dependencies? otherwise we might risk failing due to small issue at the beginning of TODO

```python
dependencies: list[Task] # or something like that
```

packages/ragbits-agents/src/ragbits/agents/tools/todo.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def get_current_task(self) -> Task | None:
+        """Get current task to work on."""
+        if self.current_index < len(self.tasks):
+            return self.tasks[self.current_index]
+        return None
+
+    def advance_to_next(self):
+        """Move to next task."""
+        self.current_index += 1
```
rk-izak: maybe introduce error recovery/dependencies here if we want them instead of rigid sequential logic? :eyes:

jakubduda-dsai: I'd recommend this for future implementation alongside with dependencies logic

PR #830, opened 2025-09-24
feat: introduce supervisor post processor

mackurzawa:


packages/ragbits-agents/src/ragbits/agents/post_processors/supervisor.py
```diff
@@ -0,0 +1,236 @@
...
+    - If the answer cannot be determined to be true because it's basic/common knowledge not covered by tools,
+      consider it as VALID
+    - If the response doesn't directly answer the user's question, flag as INVALID
+    - Be strict but fair - flag clear inconsistencies while allowing for reasonable inference
+      and interpretation of the tool results
+    """
+
+    user_prompt = (
+        "Chat History:\n"
+        "{% for message in chat_history %}\n"
```
mhordynski, line 87: W can add this logic on ChatFormat as a method. Right now ChatFormat is just an alias of dict type, but I think we could make it backward compatible.. But let's do it in separate issue. (#833)

mhordynski (approved):

Issue #834, opened 2025-10-03
feat:  agent parallel tool calling


mhordynski:
### Feature description

With parallel tool calling, LLMs will be able to call multiple tools in single turn of interaction with Ragbits agent.

For example:

- model needs to read 3 files
- includes all three tool calls in single response
- ragbits executes all tools in parallel
- results are sent back in one go

### Motivation

It is massive improvement in latency and token usage.

### Additional context

https://platform.openai.com/docs/guides/function-calling/parallel-function-calling#parallel-function-calling
https://blog.continue.dev/parallel-tool-calling/

PR #836, opened 2025-10-06
feat: agent parallel tool calling

puzzle-solver:
#834


mhordynski:
It may happen that parallel function may break existing functionality because of 2 reasons:

* sync functions are no longer executed in main thread - they need to be thread-safe and in some cases may break
* if smb implemented tools which are affected by execution order - it might change as well

because of that I am wondering if we should add a control flag to enable/disable this feature

PR #839, opened 2025-10-08
feat: long-term semantic memory

ds-michal-rdzany:
Memory tools for agents, using vector store.

It also includes an example script.


mhordynski:
Let's add a how-to tutorial to our documentation

ds-michal-rdzany:
Idk why the test has failed, I only changed a comment xd

PR #840, opened 2025-10-09
feat: conversation summary

dazy-ds:


packages/ragbits-chat/src/ragbits/chat/interface/_interface.py
```diff
@@ -371,3 +379,51 @@ async def save_feedback(
...
 
         logger.info(f"[{self.__class__.__name__}] Saving {feedback} for message {message_id} with payload {payload}")
+
+    async def generate_conversation_title(self, message: str, history: ChatFormat, context: ChatContext) -> str:
+        """
+        Default overridable generator for conversation titles.
+
+        Strategy:
+        1. cheap heuristic to detect greetings / generic messages -> "New chat"
+        2. if subclass exposes `self.llm`, call it to get a concise title
```
mhordynski: feels a little bit "magic" - maybe we should have more straightforward way to enable this feature?

dazy-ds: Yeah, I get from where you are coming from... Initially I though about using some kind of decorator for `chat` endpoint but it still didn't feel right.

On the other hand, we can for example abstract this into some kind of strategy that can be initialized and passed to the `ChatInterface` with the llm instance as a argument, but that feels too descriptive for me.

WDYT?

dazy-ds: I created `SummaryGenerator` base class along with a couple of different implementations. Now the dependency is explicit as requested and discussed.

PR #847, opened 2025-10-15
feat: #842 - oauth2 with discord

jakubduda-dsai:


packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -122,7 +124,7 @@ async def validation_exception_handler(request: Request, exc: RequestValidationE
                 content={"detail": exc.errors(), "body": exc.body},
             )
 
-    def setup_routes(self) -> None:
+    def setup_routes(self) -> None:  # noqa: PLR0915
```
dazy-ds, line 129: I think this should be refactored (extract the methods). This function grows too big and may be to complex in the future.

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -553,6 +619,150 @@ async def _handle_logout(self, request: LogoutRequest) -> JSONResponse:
                 content={"success": False, "error_message": "Internal server error"},
             )
 
+    async def _handle_oauth2_callback(  # noqa: PLR6301
```
dazy-ds, line 666: This method feels overly complex for its purpose. Generating dynamic HTML directly in the backend isn’t ideal, especially since we have a full SPA handling user flows.

I’d suggest re-architecting this so the backend only validates the code and state, then redirects to a frontend route such as /login?code=.... The frontend can then exchange that code via a dedicated API endpoint to obtain the token.

Alternatively, we could consider using an HTTP-only cookie for storing the session token, which would improve security compared to localStorage.

If we decide to keep the dynamic HTML approach for now, it would be good to at least apply consistent styling so it aligns visually with the rest of the application.

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
```diff
@@ -175,3 +179,483 @@ def cleanup_expired_tokens(self) -> int:
...
+        for backend in self.get_oauth2_backends():
+            result = await backend.authenticate_with_oauth2(oauth_credentials)
+            if result.success:
+                return result
+            if result.error_message:
+                errors.append(result.error_message)
+
+        # All backends failed
+        error_msg = "; ".join(errors) if errors else "OAuth2 authentication failed"
+        return AuthenticationResponse(success=False, error_message=error_msg)
```
dazy-ds, line 548: The naming may be confusing me (provider vs backend, as they seem to be used interchangeably), but as far as I understand the code, we do have a provider name from the callback URL. Instead of iterating over each of the backends/providers we can instead pass it directly.

This class could implement the registry pattern and register handlers for each of the provider with example syntax: `instance.registerHandler('discord', DiscordOAuth2Provider)`

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
```diff
@@ -175,3 +179,483 @@ def cleanup_expired_tokens(self) -> int:
...
+        first_backend = backends[0]
+        if hasattr(first_backend, "jwt_secret"):
+            self.jwt_secret = first_backend.jwt_secret
+            self.jwt_algorithm = first_backend.jwt_algorithm
+            self.token_expiry_minutes = first_backend.token_expiry_minutes
+        else:
+            # Fallback if first backend doesn't have JWT config
+            self.jwt_secret = get_secret_key()
+            self.jwt_algorithm = default_options.jwt_algorithm
+            self.token_expiry_minutes = default_options.token_expiry_minutes
```
dazy-ds: I think we shouldn’t naively use the first backend’s JWT configuration, since each backend may have its own (and perhaps should).

It might be better to dynamically select the correct backend and use its `validate_token` implementation, for example by routing based on a provider claim in the JWT.

typescript/ui/src/plugins/AuthPlugin/components/OAuth2Login.tsx
```diff
@@ -0,0 +1,108 @@
...
+                  fill="currentColor"
+                />
+              </g>
+              <defs>
+                <clipPath id="clip0">
+                  <rect width="71" height="55" fill="white" />
+                </clipPath>
+              </defs>
+            </svg>
```
dazy-ds, line 100: This could grow in complexity when other providers arrive. We can extract this into some kind of map outside of component which would map provider -> icon

packages/ragbits-chat/src/ragbits/chat/auth/backends.py
```diff
@@ -175,3 +179,483 @@ def cleanup_expired_tokens(self) -> int:
...
+                        "redirect_uri": self.redirect_uri,
+                    },
+                    headers={"Content-Type": "application/x-www-form-urlencoded"},
+                )
+
+                if response.status_code != 200:  # noqa: PLR2004
+                    return None
+
+                token_data = response.json()
+                return token_data.get("access_token")
```
dazy-ds, line 366: I think the `httpx.AsyncClient()` should be able to be passed as a dependency (with reasonable default), either to this method or the whole class. This would allow users to reuse the default logic with ability to add some logging, etc. without redefining the whole function.

mhordynski:
### Code review

Found 1 issue:

1. OAuth2 state parameter validation bypass allows CSRF attacks (security vulnerability)

https://github.com/deepsense-ai/ragbits/blob/89f55e9aab1cf3d3ebe224abdb934d6a47912294/packages/ragbits-chat/src/ragbits/chat/api.py#L651-L655

The state validation uses `if state and not backend.verify_state(state):` which allows the validation to be bypassed when `state` is `None` or an empty string. An attacker can craft a callback URL without a state parameter (`/api/auth/callback/discord?code=ATTACKER_CODE`) and the CSRF protection will be completely skipped. According to OAuth2 RFC 6749, the state parameter must always be validated. The correct logic should reject requests when state is missing or invalid: `if not state or not backend.verify_state(state):`.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #860, opened 2025-11-18
feat: examples for persistence strategy

MichaelMcCulloch-deepsense-ai:


mhordynski:
For now we can skip examples. we need to rework their organization as right now we have too many of them presenting just one small feature.

PR #867, opened 2025-11-26
feat: 842 - oauth discord

jakubduda-dsai:


mhordynski:
### Code review

Found 3 issues:

1. **Memory exhaustion vulnerability in OAuth2 state storage**: The `pending_states` dictionary grows indefinitely because expired states are only cleaned up when `verify_state()` is called. A malicious actor could exhaust memory by repeatedly calling `/api/auth/authorize/{provider}` without completing the OAuth2 flow, causing denial of service.

https://github.com/deepsense-ai/ragbits/blob/d8265c630d950439230af3323dc61f085431129c/packages/ragbits-chat/src/ragbits/chat/auth/backends.py#L300-L303

2. **In-memory state storage will fail in multi-instance deployments**: The OAuth2 state is stored in an in-memory dict (`pending_states`) which won't work in load-balanced environments. If the OAuth2 callback hits a different instance than the one that generated the authorization URL, state verification will fail. Consider making state storage configurable or using a shared store.

https://github.com/deepsense-ai/ragbits/blob/d8265c630d950439230af3323dc61f085431129c/packages/ragbits-chat/src/ragbits/chat/auth/backends.py#L300-L303

3. **OAuth2 callback error messages exposed in URL**: The `auth_result.error_message` is included directly in the redirect URL query parameter without sanitization or URL encoding. This could expose sensitive backend error messages in access logs, browser history, and referer headers.

https://github.com/deepsense-ai/ragbits/blob/d8265c630d950439230af3323dc61f085431129c/packages/ragbits-chat/src/ragbits/chat/api.py#L709-L713

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -302,12 +368,13 @@ async def _handle_chat_message(
                 raise HTTPException(status_code=500, detail="Chat implementation is not initialized")
 
             # Prepare chat context
-            chat_context = RagbitsAPI._prepare_chat_context(request, authenticated_user, credentials)
+            session_id = request.cookies.get("ragbits_session")
```
dazy-ds: nit: What do you think about moving `ragbits_session` to some global const for easier modification in the future if needed?

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -422,8 +487,13 @@ async def _handle_feedback(
...
         try:
-            # Validate authentication if required
-            await self._validate_authentication(credentials)
+            # Validate authentication if required using cookies
+            authenticated_user = await self.get_current_user_from_cookie(request)
+            if self.auth_backend and not authenticated_user:
+                raise HTTPException(
+                    status_code=status.HTTP_401_UNAUTHORIZED,
+                    detail="Authentication required",
+                )
```
dazy-ds: nit: This code is duplicated in a couple of handlers (and will be in more when added). What do you thing about adding this as a middleware or reusable dependency?  E.g. creating `get_authenticated_user` that would do basically the same, but reduce the duplication and allow easier usage?

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -496,34 +566,67 @@ async def _handle_feedback(
...
                     content=LoginResponse(
                         success=True,
                         user=auth_result.user if auth_result.user else None,
                         error_message=None,
-                        jwt_token=auth_result.jwt_token,
                     ).model_dump()
                 )
+
+                # Set secure HTTP-only cookie (24 hours default)
+                session_expiry_seconds = 24 * 3600
```
dazy-ds: In my opinion this should be configurable using environment variables or per auth backend. Do you think that would be beneficial?

EDIT: I see that there is `backend.session_expiry_hours` used below. Shouldn't we also use it here?

typescript/ui/src/plugins/AuthPlugin/components/AuthWatcher.tsx
```diff
@@ -1,51 +1,40 @@
...
+          login(user);
+          // Initialize the history store with user-specific key
+          initializeUserStore(user.user_id);
+        } else {
+          logout();
+        }
+      } catch (error) {
+        console.error("Failed to check session:", error);
         logout();
-        navigate("/login");
```
dazy-ds, line 24: I think we should still navigate to `/login` on error to avoid keeping users in invalid state on some URL.

typescript/ui/src/plugins/AuthPlugin/components/OAuth2Login.tsx
```diff
@@ -0,0 +1,108 @@
...
+      // Call provider-specific authorize endpoint
+      const response = await fetch(
+        `${client.getBaseUrl()}/api/auth/authorize/${provider}`,
+        {
+          method: "GET",
+          headers: {
+            "Content-Type": "application/json",
+          },
+        },
+      );
```
dazy-ds: Why aren't we using ragbits-client here? This would ensure consistency of the API calls if the interface ever changes.

typescript/ui/src/plugins/AuthPlugin/components/OAuth2Login.tsx
```diff
@@ -0,0 +1,108 @@
...
+                  fill="currentColor"
+                />
+              </g>
+              <defs>
+                <clipPath id="clip0">
+                  <rect width="71" height="55" fill="white" />
+                </clipPath>
+              </defs>
+            </svg>
+          )
```
dazy-ds: Earlier in the code I saw `OAuth2VisualConfig`. Is there a reason why we are not using it in the frontend code to define the look of OAuth buttons?

PR #871, opened 2025-11-28
feat: Add availability to pass path params

jakubduda-dsai:


typescript/@ragbits/api-client/src/types.ts
```diff
@@ -33,11 +33,19 @@ export interface StreamCallbacks<T, E = Error> {
...
 
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-export interface EndpointDefinition<Req = any, Res = any> {
+export interface EndpointDefinition<
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Req = any,
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    Res = any,
+    PathParams = never,
+    QueryParams = never,
```
dazy-ds, line 42: This doesn't seem to work. I tried creating an endpoint with defined query params (modifying `/api/feedback`) and the package didn't build.

As I understand, the goal is to allow users to define Path and Query params and be able to have autocomplete and type safety for them. If so, this needs some more work (maybe even larger definitions refactor if needed) to make it happen.

I think we should start with extending the type like so:
```
PathParams extends Record<string, unknown> | undefined = undefined
QueryParams extends Record<string, unknown> | undefined = undefined
```

Then we can fix other errors from TypeScript C:

dazy-ds (approved):

PR #872, opened 2025-11-28
fix: selecting conversation

jakubduda-dsai:


typescript/ui/src/core/stores/HistoryStore/historyStore.ts
```diff
@@ -210,6 +210,9 @@ export const createHistoryStore = immer<HistoryStore>((set, get) => ({
   actions: {
     selectConversation: (conversationId) => {
       set((draft) => {
+        if (draft.currentConversation === conversationId) {
+          return;
+        }
```
dazy-ds, line 215: I think this check should be outside of set. I am not entirely sure whether set invocation, or actual changes trigger the rerender. Let's be safe and move it outside with usage of `get()`

EDIT: I remembered that Zustand check for value equality and only triggers rerenders when value changed (or reference). That means this code is OK. Sorry for the confusion!

dazy-ds (approved):

PR #875, opened 2025-12-02
feat: Default SummaryWriter is noop when not provided

MichaelMcCulloch-deepsense-ai:


packages/ragbits-chat/src/ragbits/chat/interface/_interface.py
```diff
@@ -394,9 +392,20 @@ async def save_feedback(
...
+        Handles conversation summary generation.
+
+        If a SummaryGenerator is not configured, this logs a warning and returns None,
+        preventing log spam from unnecessary exceptions.
+        """
+        if self.summary_generator:
+            return await self.summary_generator.generate(message, history, context)
 
-        return await self.summary_generator.generate(message, history, context)
+        logging.warning(
```
mhordynski: i think we should either don't call this method at all if no summary_generator or provide a default implementation (for example beginning of first message)

mhordynski (approved):

PR #890, opened 2025-12-20
draft: batch/simulation fixups

mhordynski:


mhordynski:
it will be separated in smaller units

PR #896, opened 2025-12-31
chore: Package configuration adjustment

mkoruszowic:
This PR improves package configuration, CI/CD workflows, and resolves several dependency issues in the ragbits monorepo.

  ### Key Changes

  - **Package Configuration Refactoring**: Simplified the main `ragbits` package by referencing package extras from individual packages (e.g., `ragbits-core[chroma]`) instead of duplicating dependency definitions
  - **Build Script Enhancement**: Updated `update_nightly_versions.py` to handle optional-dependencies (extras) when updating package versions for nightly builds
  - **CI/CD Improvements**:
    - Created reusable composite action (`.github/actions/test-packages`) for shared package testing logic
    - Added dry-run mode for testing nightly builds without publishing
    - Implemented matrix strategy for testing nightly builds on Python 3.10, 3.11, and 3.12
    - Implemented matrix strategy for testing both package and source installations in shared-packages workflow
  - **Dependency Fixes (Temporary Solutions)**:
    - ⚠️ **Pinned `pyarrow==17.0.0`, `ray==2.43.0`, `docling==2.15.1`, and `torch==2.2.2`** in `.github/workflows/shared-packages.yml` as temporary workarounds
    - **Note**: These version pins are temporary solutions for test failures that occur when ragbits is installed from built packages (wheel files) rather than using `uv.lock`. When installed from builds, different versions of these libraries are resolved by the package manager, causing compatibility issues with the test suite. These pins ensure consistent behavior until the underlying dependency conflicts are properly resolved.
  - **Configuration Updates**:
    - Migrated pytest configuration from `tool.pytest` to `tool.pytest.ini_options` for backward compatibility
    - Removed uv deprecation warnings across all package `pyproject.toml` files
    - Properly linked ragbits packages in the main `pyproject.toml`
  - **Documentation**: Added ragbits-guardrails to README and documentation index

  ### Technical Details

  #### Package Configuration Refactoring

  The refactoring simplifies dependency management by having the umbrella `ragbits` package reference extras from component packages. For example, instead of:

  ```toml
  chroma = ["chromadb>=0.5.0", ... ]

  # We now use:

  chroma = ["ragbits-core[chroma]==1.4.0.dev202601010248"]
```

  This reduces duplication and ensures version consistency across the monorepo.

  ### CI/CD Testing Strategy

  The reusable composite action allows both shared-packages.yml and nightly-build.yml workflows to share the same testing logic, making the CI/CD pipeline more maintainable. The nightly build workflow now tests across multiple Python versions (3.10, 3.11, 3.12) to catch version-specific issues early.

### DIFF
--- a/release_checklist.md
+++ b/release_checklist.md
@@ -12,3 +12,19 @@
 **Status:** Currently, there is no way to install DeepEval without these dependencies.
 
 **Resolution:** These dependencies have been temporarily whitelisted in `.libraries-whitelist.txt`.
+
+### Dependency Version Pins in CI
+
+**Issue:** Test failures occur when ragbits is installed from built packages (wheel files) rather than using `uv.lock`. Different versions of the following libraries are resolved by the package manager during installation from builds, causing compatibility issues with the test suite:
+- `pyarrow`
+- `ray`
+- `docling`
+- `torch`
+
+**Status:** These are temporary workarounds until the underlying dependency conflicts are properly resolved.
+
+**Resolution:** Pinned specific versions in `.github/workflows/shared-packages.yml`:
+- `pyarrow==17.0.0`
+- `ray==2.43.0`
+- `docling==2.15.1`
+- `torch==2.2.2`

### CODE REVIEW

**Review by mhordynski — APPROVED** [review]

File: packages/ragbits/pyproject.toml [hunk 1]
```diff
@@ -41,88 +41,71 @@ dependencies = ["ragbits-document-search==1.4.0.dev202601010248", "ragbits-cli==
 
 [project.optional-dependencies]
 chroma = [
-    "chromadb>=1.0.0,<2.0.0",
```
micpst, line 44: these optional dependencies are overwritten by `scripts/update_ragbits_package.py` during release. Please update the script accordingly

pyproject.toml
```diff
@@ -60,37 +62,13 @@ members = [
...
-]
-
 [tool.pytest.ini_options]
-addopts = "--import-mode=importlib"
+norecursedirs = ".git .tox .env dist build migrations docker config notebooks research scripts examples"
+python_files = "test_*.py"
+addopts = "--import-mode=importlib -ra --showlocals --strict-markers --ignore=docs/conf.py --ignore=setup.py --ignore=ci --ignore=.eggs --tb=short " #--doctest-modules --doctest-glob=*.rst
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "function"
+pythonpath = "./"
```
micpst, line 71: nit: lets keep the old formatting, also why `--doctest-modules` and `--doctest-glob=\*.rst` are commented out?

micpst:
was playing around with the CI setup, should be a bit cleaner and faster - https://github.com/deepsense-ai/ragbits/pull/906
please take a look @mkoruszowic @mhordynski

pyproject.toml
mkoruszowic: 1. The formatting change is actually required - `ini_options` forces INI-style syntax where `addopts` must be a string (space-separated), not a TOML list
2. . The doctest options were never active before - we had both `[tool.pytest]` (ignored) and `[tool.pytest.ini_options]`, so these were dead config. When consolidated, they'd become active but cause test failures. Commented temporarily - will fix doctests in a separate PR.

packages/ragbits-chat/pyproject.toml
```diff
@@ -44,8 +44,8 @@ sql = [
     "sqlalchemy>=2.0.39,<3.0.0",
 ]
 
-[tool.uv]
-dev-dependencies = [
+[dependency-groups]
```
micpst: we don't need `dev` dependencies per package, since we have them defined in the root `pyproject.toml`, we can safely delete them, wdyt?

mkoruszowic: I think so

mkoruszowic:
@micpst do you think we need to mark to use those pinned libraries in documentation

micpst:
nice to have, but since we don't recommend installing all extras in the docs, I'd leave it as is

PR #905, opened 2026-01-12
feat(chat): implement file upload ingestion support

MichaelMcCulloch-deepsense-ai:
- Add `upload_handler` to `ChatInterface` to support handling file uploads
- Add `/api/upload` endpoint to `RagbitsAPI`
- Add `uploadFile` method to TypeScript `RagbitsClient`
- Add file upload button to `PromptInput` component
- Add example `examples/chat/upload_chat.py`
- Add unit tests for upload functionality"

### DIFF
--- a/typescript/ui/__tests__/integration/intergation.test.tsx
+++ b/typescript/ui/__tests__/integration/intergation.test.tsx
@@ -329,6 +329,7 @@ describe("Integration tests", () => {
         }
         return Promise.resolve(new Response("{}", { status: 200 }));
       });
+      const originalFetch = global.fetch;
       global.fetch = fetchSpy;
 
       const WrappedInput = () => (
@@ -382,6 +383,8 @@ describe("Integration tests", () => {
         const headers = options?.headers as Record<string, string>;
         expect(headers["Content-Type"]).toBeUndefined();
       });
+
+      global.fetch = originalFetch;
     });
   });
 


typescript/@ragbits/api-client/src/index.ts
```diff
@@ -104,6 +104,46 @@ export class RagbitsClient {
...
+
+        if (!response.ok) {
+            const error = await response.json().catch(() => ({}))
+            throw new Error(
+                error.detail || `Upload failed with status ${response.status}`
+            )
+        }
+
+        return response.json()
+    }
```
dazy-ds: I’m not entirely convinced that defining `uploadFile` in this way is the best approach. It seems that this functionality could likely be handled by the existing `makeRequest` base function, possibly with a small adjustment to how default headers are applied.

Additionally, it would be preferable to avoid hardcoding URLs in this form, as consumers of the library may want the flexibility to override them. One possible alternative would be to extend `BaseApiEndpoints` with a new entry, for example:
```typescript
    '/api/upload': EndpointDefinition<
        FormData,
        { status: string; filename: string }
    >
```

This would allow us to use a preferred approach in React without using the `client` directly (which is on lower abstraction layer), such as:
```typescript
const { call, data, isLoading: isUploading } = useRagbitsCall('/api/upload');
...
const formData = new FormData();
formData.append("file", file);
await uploadFile({
  body: formData,
  method: "POST",
});
```

Implementing this would require some updates to `makeRequest` and `_makeRequest` so they can detect when the request body is a `FormData` instance and avoid automatically stringifying it or setting the `application/json` Content-Type header.

Later this approach could also serve as a clear and reusable pattern for handling file uploads for user defined endpoints.

typescript/ui/src/core/components/inputs/PromptInput/PromptInput.tsx
```diff
@@ -55,6 +58,9 @@ const PromptInput = ({
   const { isCaretInFirstLine, isCaretInLastLine } =
     useCaretLogicalLineDetection();
   const quickMessageIndex = useRef(Math.max(quickMessages.length - 1, 0));
+  const fileInputRef = useRef<HTMLInputElement>(null);
+  const { client } = useRagbitsContext();
+  const [isUploading, setIsUploading] = useState(false);
```
dazy-ds: This should no longer be necessary once we switch to using the `useRagbitsClient` hook and approach mentioned above.

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -226,6 +230,10 @@ async def feedback(
             ) -> JSONResponse:
                 return await self._handle_feedback(feedback_request, request)
 
+            @self.app.post("/api/upload", response_class=JSONResponse)
+            async def upload_file(file: UploadFile) -> JSONResponse:
+                return await self._handle_file_upload(file)
+
```
mkoruszowic: Why duplicate this? Can't it be outside the if statement since it applies to both cases?

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -585,6 +593,31 @@ async def _handle_feedback(self, feedback_request: FeedbackRequest, request: Req
...
+        """
+        Handle file upload requests.
+
+        Args:
+            file: The uploaded file.
+
+        Returns:
+            JSONResponse with status.
+        """
+        if not self.chat_interface.upload_handler:
```
mkoruszowic: More robust IMO is to explicitly check for None

packages/ragbits-chat/src/ragbits/chat/interface/_interface.py
```diff
@@ -211,6 +211,15 @@ class ChatInterface(ABC):
     ui_customization: UICustomization | None = None
     history_persistence: HistoryPersistenceStrategy | None = None
     summary_generator: SummaryGenerator = HeuristicSummaryGenerator()
+    upload_handler: Callable[[Any], Any] | None = None
```
mkoruszowic: Should it always take `UploadFile`  and not `Any`?

packages/ragbits-chat/src/ragbits/chat/interface/_interface.py
```diff
@@ -211,6 +211,15 @@ class ChatInterface(ABC):
...
+    """
+    Optional handler/callback for file uploads.
+    The handler should be an async method that accepts a single argument `file` (UploadFile).
+    Example:
+        async def upload_handler(self, file: UploadFile):
+            content = await file.read()
+            # process content
+    """
```
mkoruszowic: I think it would be better to have it in the class docstring under  attributes e.g.:
```
"""
Base interface for chat implementations.
.
.
.
* State updates: Updates to the conversation state

Attributes:
    feedback_config: Configuration for user feedback collection.
    ...
    upload_handler: Optional async callback for handling file uploads.
        Should accept an UploadFile parameter and return None.
        
        Example::
        
            async def upload_handler(self, file: UploadFile) -> None:
                content = await file.read()
                # process content
"""
````

File: typescript/@ragbits/api-client/src/index.ts [hunk 7]
```diff
@@ -134,8 +134,21 @@ export class RagbitsClient {
...
+                    'Content-Type' in requestOptions.headers
+                ) {
+                    delete (requestOptions.headers as Record<string, string>)[
+                        'Content-Type'
+                    ]
+                }
+            } else {
+                requestOptions.body =
+                    typeof body === 'string' ? body : JSON.stringify(body)
+            }
```
dazy-ds, line 155: I've performed some tests and the current implementation doesn't seem to work. We are passing empty `requestOptions.headers` object to the `_makeRequest` method, which applies default headers. The default headers aren't overridden because `requestOptions.headers` field doesn't contain any `Content-Type` definition.

I think we should set default headers in the `_makeRequest` conditionally. Something like:
```typescript
        const defaultHeaders: Record<string, string> =
            typeof options.body === 'string' // we guess that it's JSON
                ? {
                      'Content-Type': 'application/json',
                  }
                : {}
 ```
 and also keep this part here.

MichaelMcCulloch-deepsense-ai:
Thanks!

PR #906, opened 2026-01-12
refactor: ci tests

micpst:


.github/workflows/shared-packages.yml
```diff
@@ -181,23 +208,23 @@ jobs:
...
-      - name: Run Tests With Coverage (source)
-        if: matrix.install-method == 'source'
         run: |
-          uv run coverage run -m pytest --postgresql-password postgres -v -p no:warnings --junitxml=report.xml
-          uv run coverage report
-          uv run coverage xml
+          uv sync --only-group dev
+          uv pip install --find-links dist \
+            "$(ls dist/ragbits-*.whl)[a2a,azure,chroma,cli,fastembed,fastembed-gpu,gcs,google-drive,hf,local,logfire,mcp,openai,otel,pgvector,promptfoo,qdrant,ray,relari,s3,sql,unstructured,weaviate]" \
+            docling==2.15.1 torch==2.2.2 pyarrow==17.0.0 ray==2.43.0
```
mkoruszowic, line 220: Do we also need to add a note here stating that this is a temporary workaround, given that we’ll probably be updating the documentation anyway and it will be self explanatory that `docling==2.15.1 torch==2.2.2 pyarrow==17.0.0 ray==2.43.0` is here?

Issue #910, opened 2026-01-20
feat: hooking agent mechanism as a more general replacement for PostProcessors


mkoruszowic:
### Feature description

Hooks should exte PostProcessors fucntionality of using them not only at the end of the run but also at the beggining or  due to some events like tools.  They shoudl rpvoide clear mechanism enalbing user to attach specific hooks that will be trgier due to some event  in specific part of the code.. Possible hooks type are mentioned here https://platform.claude.com/docs/en/agent-sdk/hooks#available-hooks but at the beggining we can focus on hooks:
- pre run 
- post run
- pre tool
- post tool
Other importnant think to mention or consider is supporting both streaming and non streaming run.

### Motivation

Not clear way of handling input/output or even tool check. No documentented way of introducing guradrails to agent as a validation step. Origantes form: https://github.com/deepsense-ai/ragbits/issues/891 as a continuation of the idea of otput fucntion placed as hooks for different evetns

### Additional context

Equivalent implementations in other frameworks:
- https://platform.claude.com/docs/en/agent-sdk/hooks#common-use-cases
- https://openai.github.io/openai-agents-python/guardrails/#tool-guardrails
- https://ai.pydantic.dev/output/#output-functions

Ground truth use case:
...

PR #914, opened 2026-01-23
feat: Introducing hook mechanism

mkoruszowic:
## Summary                                                                                                                                                                                                               
                                                                                                                                                                                                                           
  Introduces a **hooks system** for agent lifecycle event interception.
  Related issue: https://github.com/deepsense-ai/ragbits/issues/910                                                                                                                                        
                                                                                                                                                                                                                           
  ### Implemented (from issue requirements)                                                                                                                                                                                
                                                                                                                                                                                                                           
  | Hook Type | Status | Description |                                                                                                                                                                                     
  |-----------|--------|-------------|                                                                                                                                                                                     
  | **pre-tool** | ✅ Done | Execute logic before tool invocation |                                                                                                                                                         
  | **post-tool** | ✅ Done | Execute logic after tool completion |                                                                                                                                                         
  | pre-run | 📋 Separate PR | Execute before agent starts |                                                                                                                                                               
  | post-run | 📋 Separate PR | Execute after agent finishes |                                                                                                                                                             
                                                                                                                                                                                                                           
  ### Target Use Cases (from issue)                                                                                                                                                                                        
                                                                                                                                                                                                                           
  | Use Case | Status |                                                                                                                                                                                                    
  |----------|--------|                                                                                                                                                                                                    
  | Tool confirmation (pre-tool) | ✅ `requires_confirmation_hook()` |                                                                                                                                                     
  | Tool access control (pre-tool) | ✅ Via `decision="deny"` |                                                                                                                                                            
  | Input validation (pre-tool) | ✅ Argument modification + chaining |                                                                                                                                                    
  | Output validation (post-tool) | ✅ Output modification + chaining |                                                                                                                                                    
                                                                                                                                                                                                                           
  ### Key Features                                                                                                                                                                                                         
                                                                                                                                                                                                                           
  - **Generic `Hook` class** - Type-safe callbacks with `Hook[InputT, OutputT]`                                                                                                                                            
  - **`HookManager`** - Registration, priority ordering, and chained execution                                                                                                                                             
  - **Decision flow** - `pass` / `ask` / `deny` for pre-tool hooks                                                                                                                                                         
  - **Hook-based confirmation** - Replaces `@requires_confirmation` decorator with `requires_confirmation_hook()` factory                                                                                                  
  - **`ConfirmationRequest` embedding** - `PreToolOutput` contains full request when `decision="ask"`                                                                                                                      
  - **UI integration** - Frontend handles confirmation dialogs via `confirmed_hooks` context                                                                                                                               
                                                                                                                                                                                                                           
  ### Example Usage                                                                                                                                                                                                        
                                                                                                                                                                                                                           
  ```python                                                                                                                                                                                                                
  from ragbits.agents import Agent                                                                                                                                                                                         
  from ragbits.agents.hooks import requires_confirmation_hook                                                                                                                                                              
                                                                                                                                                                                                                           
  agent = Agent(                                                                                                                                                                                                           
      llm=llm,                                                                                                                                                                                                             
      tools=[delete_file, create_file],                                                                                                                                                                                    
      hooks=[requires_confirmation_hook(tools=["delete_file"])]                                                                                                                                                            
  )      
```

### DIFF
--- a/examples/agents/hooks/agent_output_logging.py
+++ b/examples/agents/hooks/agent_output_logging.py
@@ -1,0 +1,72 @@
+"""
+Example demonstrating a post-tool hook for logging outputs from agent tools.
+
+This shows how to log the output returned by downstream agent tools.
+Setup: 1 parent agent with 2 expert child agents as tools.
+"""
+
+import asyncio
+
+from ragbits.agents import Agent
+from ragbits.agents.hooks import EventType, Hook, PostToolInput, PostToolOutput
+from ragbits.core.llms import LiteLLM
+
+
+async def log_agent_output(input_data: PostToolInput) -> PostToolOutput:
+    """Log output from agent tools after they complete."""
+    output = input_data.output
+
+    if isinstance(output, dict):
+        output = output.get("content", output)
+
+    print(f"\n{'='*60}\n[{input_data.tool_call.name}] output:\n{'-'*60}\n{output}\n{'='*60}")
+    return PostToolOutput(output=input_data.output)
+
+
+async def main() -> None:
+    """Run the example."""
+    llm = LiteLLM("gpt-4o-mini")
+
+    # Child agent 1: Diet expert
+    diet_agent = Agent(
+        name="diet_expert",
+        description="A nutrition expert who provides diet plans and healthy eating advice",
+        llm=llm,
+    )
+
+    # Child agent 2: Fitness coach
+    fitness_agent = Agent(
+        name="fitness_coach",
+        description="A personal trainer who creates workout routines and exercise plans",
+        llm=llm,
+    )
+
+    # Hook to log outputs from both agent tools
+    hook = Hook(
+        event_type=EventType.POST_TOOL,
+        callback=log_agent_output,
+        tool_names=["diet_expert", "fitness_coach"],
+        priority=1,
+    )
+
+    # Parent agent with both expert agents as tools
+    parent_agent = Agent(
+        name="health_assistant",
+        llm=llm,
+        tools=[
+            (diet_agent, {"name": "diet_expert"}),
+            (fitness_agent, {"name": "fitness_coach"}),
+        ],
+        hooks=[hook],
+    )
+
+    # Query that should trigger both agent tools
+    query = "I want to lose 10kg in 3 months. Can you help me with a plan?"
+    print(f"Query: {query}\n")
+
+    response = await parent_agent.run(query)
+    print(f"Final Response:\n{response.content}")
+
+
+if __name__ == "__main__":
+    asyncio.run(main())
--- a/examples/agents/hooks/validation_and_sanitization.py
+++ b/examples/agents/hooks/validation_and_sanitization.py
@@ -137,7 +137,9 @@ async def main() -> None:
 
     hooks = [
         Hook(event_type=EventType.PRE_TOOL, callback=validate_email, tool_names=["send_notification"], priority=10),
-        Hook(event_type=EventType.PRE_TOOL, callback=sanitize_email_domain, tool_names=["send_notification"], priority=20),
+        Hook(
+            event_type=EventType.PRE_TOOL, callback=sanitize_email_domain, tool_names=["send_notification"], priority=20
+        ),
         Hook(event_type=EventType.POST_TOOL, callback=mask_sensitive_data, tool_names=["search_user"], priority=10),
         Hook(event_type=EventType.POST_TOOL, callback=log_notification, tool_names=["send_notification"], priority=10),
     ]

### CODE REVIEW

File: packages/ragbits-agents/src/ragbits/agents/hooks/confirmation.py [hunk 1]
```diff
@@ -0,0 +1,48 @@
...
+Helper functions for creating common hooks.
+
+This module provides factory functions for creating commonly used hooks.
+"""
+
+from ragbits.agents.hooks.base import Hook
+from ragbits.agents.hooks.types import EventType, PreToolInput, PreToolOutput
+
+
+def requires_confirmation_hook(tools: list[str] | None = None, priority: int = 1) -> Hook[PreToolInput, PreToolOutput]:
```
puzzle-solver: IMO a better name would be `create_confirmation_hook`, it is not clear why `requires` is there.

packages/ragbits-agents/src/ragbits/agents/hooks/confirmation.py
```diff
@@ -0,0 +1,48 @@
...
+    """
+    Create a hook that requires user confirmation before tool execution.
+
+    This replaces the @requires_confirmation decorator with a hook-based approach.
```
puzzle-solver: I think we shouldn't reference previous implementation in the docs, since it has never been released anyway.

packages/ragbits-agents/src/ragbits/agents/hooks/confirmation.py
```diff
@@ -0,0 +1,48 @@
...
+        tools: List of tool names to require confirmation for. If None, applies to all tools.
+        priority: Hook priority (default: 1, runs first)
+
+    Returns:
+        Hook configured to require confirmation
+
+    Example:
+        ```python
+        from ragbits.agents import Agent
+        from ragbits.agents.hooks.helpers import requires_confirmation_hook
```
puzzle-solver: It should be ` from ragbits.agents.hooks.confirmation import requires_confirmation_hook`

packages/ragbits-agents/src/ragbits/agents/hooks/base.py
```diff
@@ -0,0 +1,80 @@
...
+from dataclasses import dataclass
+from typing import Generic, TypeVar
+
+from ragbits.agents.hooks.types import EventType, HookEventIO
+
+HookInputT = TypeVar("HookInputT", bound=HookEventIO)
+HookOutputT = TypeVar("HookOutputT", bound=HookEventIO)
+
+
+@dataclass
```
puzzle-solver: I am not sure if making it a dataclass makes sense. This class is not really just for storing data. I would suggest making it a regular class.

packages/ragbits-agents/src/ragbits/agents/hooks/confirmation.py
```diff
@@ -0,0 +1,48 @@
...
+def requires_confirmation_hook(tools: list[str] | None = None, priority: int = 1) -> Hook[PreToolInput, PreToolOutput]:
+    """
+    Create a hook that requires user confirmation before tool execution.
+
+    This replaces the @requires_confirmation decorator with a hook-based approach.
+    The hook returns "ask" decision, which causes the agent to yield a ConfirmationRequest
+    and wait for user approval/decline.
+
+    Args:
+        tools: List of tool names to require confirmation for. If None, applies to all tools.
```
puzzle-solver: I think that `tools` ma be slightly misleading here, since we are passing tool names, not tools themselves.

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -212,9 +210,9 @@ class AgentRunContext(BaseModel, Generic[DepsT]):
     """Whether to stream events from downstream agents when tools execute other agents."""
     downstream_agents: dict[str, "Agent"] = Field(default_factory=dict)
     """Registry of all agents that participated in this run"""
-    confirmed_tools: list[dict[str, Any]] | None = Field(
-        default=None,
-        description="List of confirmed/declined tools from the frontend",
+    confirmed_hooks: list[dict[str, Any]] = Field(
+        default_factory=list,
+        description="List of confirmed/declined hooks. Each entry has 'confirmation_id' and 'confirmed' (bool)",
```
puzzle-solver: Isn't it tools that we confirm/decline, not hooks?

puzzle-solver (commented):
Overall, I think that the idea makes sense. I had a few nitpicks here and there.

Also, I am slightly unsure of the interface (for example why `Hook` with event_type instead of just `PreToolHook` and `PostToolHook` with method `execute` already having appropriate arguments:

```
class PreToolHook(Hook):
    def execute(tool_call: ToolCall) -> PreToolOutput:
        return self.callback(tool_call)
```

, which would be slightly more explicit.

packages/ragbits-agents/src/ragbits/agents/hooks/confirmation.py
mkoruszowic: Right good catch, I missed that

packages/ragbits-agents/src/ragbits/agents/_main.py
mkoruszowic: In general, it's currently multi-acceptance. Multiple hooks targeting the same tool can issue an "ask," hence the confirm_hooks. This is based on the Claude SDK because I found this approach more universal: https://platform.claude.com/docs/en/agent-sdk/hooks#permission-decision-flow. But maybe `tool_confirmations` would be a better name as hooks suggest it can be applied to all of those types

Issue #919, opened 2026-01-27
feat: streaming events from tools


puzzle-solver:
### Feature description

I propose to implement a feature view streaming tools directly from tools. In this way, we unify tools and agents used as tools, which already support streaming downstream events.

Example usage:
```
def send_notification(user_id: str, message: str) -> Generator[NotificationResponse | ToolReturn]:
    user = get_user(user_id)
    if user is not None:
        yield NotificationResponse(content=NotificationContent(user_id=user.id, message=message))
        yield ToolReturn(value={"success": True}, metadata={"user": user.to_dict()})
    else:
        yield ToolReturn(value={"success": False, "error": "User does not exist"})
```

Here, we use a `ToolReturn` to indicate a return value that we need to pass to the LLM. We assume that the generator yields only one `ToolReturn`. The `NotificationResponse` will be be streamed directly from an agent (assuming an appropriate option is set to True).

In general, we assume the tools passed to an agent to be one of the following:
1. Raw Tool object
2. Function
3. Async function
4. **Generator function**
5. **Async generator function**
6. Agent

This feature proposes adding support for generators and async generators as tools.

### Motivation

A useful agent needs to be able to interact with the world. The main way for an agent to interact with the world is through tools. In some cases, an agent needs to send an event to another system to indicate something. This may be, for example, an add an item to the cart in the mobile application, with which the user interacts with via chat. To handle it right now, we need to have a complicated architecture, in which we process all of the events in the `ChatInterface`, match the type of the tool call (`if result.name == "modify_cart:`), extract the output of the tool and send it. Instead, we may want to do it directly in a tool.

puzzle-solver:
Already implemented, merged and released in version 1.5

PR #921, opened 2026-01-28
feat: streaming events from tools

puzzle-solver:
Adds support for streaming events directly from tools, when they are a generator or async generator functions. Agent streams all events from tools, and the tool outputs are defined by objects of type `ToolReturn`. 

Detailed description of a feature: #919 

Limitations:
* if the generator is synchronous, the iteration through it is blocking even when `parallel_tool_calling` is set to True (there is no mechanism to move it to a thread that can be executed in parallel with other tools)


packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -405,15 +413,14 @@ def __init__(
         self.description = description
         self.tools = []
         for tool in tools or []:
-            if isinstance(tool, tuple):
-                agent, kwargs = tool
-                self.tools.append(Tool.from_agent(agent, **kwargs))
-            elif isinstance(tool, Agent):
+            if isinstance(tool, Agent):
```
mkoruszowic, line 427: If you want to exclude creation from tuple, you need to adjust `examples/agents/downstream_agents_streaming.py` and docs because this form is supported there

puzzle-solver: Fixed

puzzle-solver:
> What about extending `How-To: Define and Use Agents with Ragbits` by adding examples of new tool types or practical usage scenarios?

Good idea, I added an example and a how-to. I think that the example illustrates quite well the use case that I had in mind when implementing this.

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -1058,8 +1069,10 @@ async def _execute_tool(
                 async for tool_output_event in self._process_tool_output(tool_output, tool, context):
                     if isinstance(tool_output_event, ToolReturn):
                         tool_return = tool_output_event
-                    else:
+                    elif isinstance(tool_output_event, ToolEvent):
                         yield tool_output_event
+                    else:
+                        yield ToolEvent(content=tool_output_event)
```
mkoruszowic: Is it intentional to skip somehow post tool hook by already yielding the `ToolEvent`

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -1058,8 +1069,10 @@ async def _execute_tool(
                 async for tool_output_event in self._process_tool_output(tool_output, tool, context):
                     if isinstance(tool_output_event, ToolReturn):
                         tool_return = tool_output_event
-                    else:
```
mkoruszowic: What happens if we iterate and receive `ToolReturn` multiple times? In that case, only the last one will actually be used.  Comment to this part (my bad):
```
                async for tool_output_event in self._process_tool_output(tool_output, tool, context):
                    if isinstance(tool_output_event, ToolReturn):
                        tool_return = tool_output_event
                    elif isinstance(tool_output_event, ToolEvent):
                        yield tool_output_event
                    else:
                        yield ToolEvent(content=tool_output_event)
```

examples/agents/stream_events_from_tools.py
mkoruszowic, line 1: Since it is in `examples/agents/` would it be better to use just `Agent` instead of `ChatInterface`? You could just move it to the docs and replace it here with the `Agent`

packages/ragbits-agents/src/ragbits/agents/_main.py
mkoruszowic: It is outdated because I marked wrong place it is not resolved

packages/ragbits-agents/src/ragbits/agents/_main.py
mkoruszowic: Ok in docs there is an info about. Maybe additional warning would be beneficial if someone forgot about it

puzzle-solver: Yes, only one ToolReturn is allowed. I added a warning when tools yield more than one ToolReturn, it is the only way to handle it I think

examples/agents/stream_events_from_tools.py
puzzle-solver: In this script there is a dependency of the `ragbits-chat` as well, so the examples will run no problem. However, I added a simpler setup of agent with streaming tools to demonstrate the usage explicitly. I left the streaming events to the Ragbits UI in a separate file, because I think it is a cool demonstration what a streaming tool can do.

mkoruszowic (approved):

PR #922, opened 2026-01-29
fix: proper trace context manager closing for streaming response

sleter:


packages/ragbits-core/src/ragbits/core/llms/base.py
```diff
@@ -244,6 +244,9 @@ async def __anext__(self) -> T:
                     self.metadata: LLMResponseWithMetadata = item
                     if item.usage:
                         self.usage += item.usage
+                    # Exhaust generator to let trace context managers close properly
+                    async for _ in self._generator:
+                        pass
```
mkoruszowic, line 249: Isn't  `await self._generator.aclose()` better?

sleter: `aclose` throws `GeneratorExit` that is later caugh in `packages/ragbits-core/src/ragbits/core/audit/traces/base.py60:90`

that causes that these lines are not executed properly:
```python
span = self._spans.get().pop()
self.stop(outputs=vars(outputs), current_span=span)
```

PR #925, opened 2026-02-02
fix: fixed post tool hooks behavior with tool output

puzzle-solver:


examples/agents/hooks/validation_and_sanitization.py
```diff
@@ -89,10 +90,10 @@ async def sanitize_email_domain(input_data: PreToolInput) -> PreToolOutput:
 async def mask_sensitive_data(input_data: PostToolInput) -> PostToolOutput:
     """Mask sensitive information in user search results."""
     if input_data.tool_call.name != "search_user":
-        return PostToolOutput(output=input_data.output)
+        return PostToolOutput(tool_return=input_data.tool_return)
 
     if isinstance(input_data.output, dict) and "ssn" in input_data.output:
```
mkoruszowic: This also needs to be adjusted to make the example running

packages/ragbits-agents/src/ragbits/agents/_main.py
```diff
@@ -1051,8 +1050,8 @@ async def _execute_tool(
             id=tool_call.id,
             name=tool_call.name,
             arguments=tool_call.arguments,
-            result=tool_return.value,
-            metadata=tool_return.metadata,
+            result=post_tool_output.tool_return.value if post_tool_output.tool_return else None,
+            metadata=post_tool_output.tool_return.metadata if post_tool_output.tool_return else None,
```
mkoruszowic, line 1054: Why the if statement? Metadata  will by default be None if nothing is passed and `value` should always be smth

examples/agents/hooks/validation_and_sanitization.py
puzzle-solver: Adjusted, thanks for the catch!

packages/ragbits-agents/src/ragbits/agents/_main.py
puzzle-solver: This is primarily because of the typing. `execute_post_tool` returns `PostToolOutput`, with tool_return that can potentially be `None` (if there was an exception during running the tool). In practice, it is not possible to have a `tool_return` as None at this point, because we raise the exception after running all the post hooks anyway. However, mypy will complain nevertheless.

mkoruszowic: ok, I see. On my hook branch I removed the error field completely and execute exception even before the hook that's why it is working. Nevertheless let's leave it as it is I will change it in my PR then

Issue #937, opened 2026-02-12
feat: on-event hook for streaming


mkoruszowic:
### Feature description

**on-event** hook -  executed while streaming handling different events/StreamingPostProcessor replacement

### Motivation

In parent issue

### Additional context

In parent issue

PR #939, opened 2026-02-16
feat: on-event hook

mkoruszowic:
### Summary                                                                                                                                                                                                              
                                                                              
  - **Replaced hook IO wrapper types with direct Callable aliases** — removed `PreToolInput`/`PreToolOutput` and similar Protocol-based wrapper classes in favor of simple `Callable` type aliases (`PreToolCallback`,     
  `PostToolCallback`, `PreRunCallback`, `PostRunCallback`, `OnEventCallback`). Hook callbacks now receive and return domain objects directly (e.g. `ToolCall`, `ToolReturn`, `AgentResult`) instead of being               
  wrapped/unwrapped.
  - **Extended `ToolCall` with `decision`/`reason` fields** — pre-tool hooks now return a `ToolCall` with an optional `decision` (`pass`/`deny`/`ask`) and `reason`, eliminating the need for a separate output wrapper
  type.
  - **Added `ON_EVENT` hook type** — replaces `StreamingPostProcessor` with a new hook event that intercepts streaming events (`str` chunks, `ToolCall`, `ToolCallResult`) during `run_streaming()`. Hooks can modify,
  suppress, or pass through events. Hooks chain in priority order, and returning
  `None` suppresses the event.
  - **Made `options` and `context` non-optional in `Agent.run()`** — internally these are always created, so the hook signatures and internal plumbing now treat them as required rather than `Optional`.
  - **Made `Hook` generic on callback type** — `Hook[CallbackT]` with `@overload` on `get_hooks()` returns `list[Hook[PreToolCallback]]` etc., eliminating all `# type: ignore[call-arg]` comments from the hook execution
  methods.


Related  issue: https://github.com/deepsense-ai/ragbits/issues/937

### DIFF
--- a/packages/ragbits-agents/src/ragbits/agents/hooks/manager.py
+++ b/packages/ragbits-agents/src/ragbits/agents/hooks/manager.py
@@ -290,7 +290,6 @@ class HookManager(Generic[LLMClientOptionsT, PromptInputT, PromptOutputT]):
                     if awaited is not None:
                         yield awaited
 
-
         for hook in self.get_hooks(EventType.ON_EVENT):
             generator = apply_hook_to_stream(generator, hook)
 
--- a/packages/ragbits-agents/src/ragbits/agents/_main.py
+++ b/packages/ragbits-agents/src/ragbits/agents/_main.py
@@ -1106,7 +1106,7 @@ class Agent(
                 id=tool_call.id,
                 name=tool_call.name,
                 arguments=tool_call.arguments,
-                result="⏳ Awaiting user confirmation"
+                result="⏳ Awaiting user confirmation",
             )
             return
 


packages/ragbits-agents/src/ragbits/agents/hooks/manager.py
```diff
@@ -91,85 +115,81 @@ async def execute_pre_tool(
...
             confirmation_id = hashlib.sha256(confirmation_id_str.encode()).hexdigest()[:CONFIRMATION_ID_LENGTH]
 
-            # Create input with current state (chained from previous hook)
-            hook_input = PreToolInput(
-                tool_call=tool_call.model_copy(update={"arguments": current_arguments}),
-            )
+            result: ToolCall = await hook.callback(current_tool_call)
 
-            result: PreToolOutput = await hook.execute(hook_input)
+            # Validation (moved from PreToolOutput.__post_init__)
```
puzzle-solver: Comment to delete

examples/agents/hooks/agent_output_logging.py
puzzle-solver: Here, we need to modify it to just `tools=[diet_agent, fitness_agent]`. The names are simply taken form the `name` attribute, no need to pass the tuple (hence I removed that option from the api, as it looks kind of weird).

packages/ragbits-agents/src/ragbits/agents/hooks/manager.py
```diff
@@ -259,25 +252,39 @@ async def execute_post_run(
...
+            Processed streaming events (potentially modified by hooks)
+        """
+        hooks = self.get_hooks(EventType.ON_EVENT)
+
+        async for event in generator:
+            current_event: StreamingEvent | None = event
+            for hook in hooks:
+                if current_event is None:
+                    break
+                current_event = await hook.callback(current_event)
```
puzzle-solver: How about extending it to allow a hook to yield multiple events based on a single one? Could be potentially useful and it seems easy to extend here.

docs/how-to/agents/use_hooks.md
```diff
@@ -43,15 +44,15 @@ agent = Agent(
 The `Hook` constructor accepts:
 
 - **`event_type`** — one of `EventType.PRE_TOOL`, `POST_TOOL`, `PRE_RUN`, or `POST_RUN`
```
puzzle-solver: `ON_EVENT` as well

examples/agents/hooks/agent_output_logging.py
mkoruszowic: Right, I forgot about about it

PR #971, opened 2026-04-14
chore: add docker setup for local testing

aholka:
Use testcontainers library to satisfy integration test dependencies (postgresql with pgvector and weaviate) automatically.


CONTRIBUTING.md
```diff
@@ -34,8 +34,31 @@ We use `pytest` for testing. To run the tests, simply run:
...
+You can check that services are healthy with:
+
+```bash
+docker compose ps
+```
+
+And stop them with:
+
+```bash
+docker compose down
```
mkoruszowic: Better to use `docker compose down -v` to also remove volumes, otherwise postgres data persists and schema changes / init.sql updates won't take effect.

pyproject.toml
```diff
@@ -69,6 +69,9 @@ addopts = "--import-mode=importlib -ra --showlocals --strict-markers --ignore=do
 asyncio_mode = "auto"
 asyncio_default_fixture_loop_scope = "function"
 pythonpath = "./"
+postgresql_user = "postgres"
+postgresql_password = "postgres"
+postgresql_dbname = "tests"
```
mkoruszowic: The `postgresql_password` ini option doesn't actually work `postgresql_noproc` ignores it when making its initial version-check connection (`executor_noop.py:70`). I've checked it tests fail without  `--postgresql-password postgres` on the CLI, even with this setting present. IMO remove these three lines (they give a false sense of configuration) and document that `--postgresql-password` postgres is still required. CI already passes it via CLI (line 232 of `shared-packages.yml`).

mkoruszowic (changes requested):

pyproject.toml
mkoruszowic: My fault. It is working for `uv run pytest`. It only need passing an explicit password in case of running those test individually

CONTRIBUTING.md
aholka: That's a good catch, thank you. But I've proposed a different solution now - use testcontainers dependency instead of manually define docker-compose file.

packages/ragbits-core/tests/integration/vector_stores/conftest.py
```diff
@@ -0,0 +1,40 @@
+from collections.abc import AsyncGenerator
+
+import asyncpg
+import pytest
+from testcontainers.postgres import PostgresContainer
+from testcontainers.weaviate import WeaviateContainer
+
+POSTGRES_IMAGE = "pgvector/pgvector:pg17"
```
mkoruszowic: the old CI config had 0.8.0-pg17. Weaviate is still pinned, so this looks accidental. Could you pin it back to avoid surprises from upstream releases?

packages/ragbits-core/tests/integration/vector_stores/conftest.py
mkoruszowic, line 1: the postgres fixture does `DROP SCHEMA public CASCADE` per request, but weaviate is session-scoped with no reset. If a test fails and leaves an index behind, the next one could pick up stale data. Could you check whether the existing tests handle their own cleanup, or add a teardown?

mkoruszowic (commented):
PR description mentions "added a Docker compose file" but there's no longer a compose file, it's all testcontainers. Mind updating the body so it matches what actually landed?

packages/ragbits-core/tests/integration/vector_stores/conftest.py
aholka: I've added weaviate cleanup per each test, so postgres and weaviate are handled the same way now

mkoruszowic (approved):
LGTM

pskwiercz (approved):
LGTM

PR #973, opened 2026-04-20
feat: make litellm dependency optional

### LINKED ISSUE #967: feat: make LiteLLM dependency optional
### Feature description

LiteLLM should be removed from default paths.
OpenAI / Anthropic / Google Gemini dedicated classes should be created, with OpenAI as the default option

- dependency on LiteLLM can stil exist, ~ 80% should work without it
- rerankers (ragbits-document-search) can be left as-is
- add new LLM classes
- add new embedder classes
- core functionality:
  - focus on supporting OpenAI / Anthropic / Google Gemini
  - support Ollama / VLLM (hosted locally)

### Motivation

LiteLLM should not be a mandatory dependency

### Additional context

_No response_

aholka:
Breaking change: LiteLLM becomes an optional dependency.
New LLM implementations: OpenAILLM, AnthropicLLM, GeminiLLM
New embedder implementations: OpenAIEmbedder, GeminiEmbedder.

note: this branch contains changes from add-docker-containers-for-integration-testing branch which hasn't been merged to main yet


packages/ragbits-core/src/ragbits/core/embeddings/dense/openai.py
```diff
@@ -0,0 +1,145 @@
...
+    HAS_OPENAI = True
+except ImportError:
+    HAS_OPENAI = False
+
+
+class OpenAIEmbedderOptions(Options):
+    """
+    Dataclass that represents available call options for the OpenAIEmbedder client.
+    Each of them is described in the
+    [OpenAI API documentation](https://platform.openai.com/docs/api-reference/embeddings/create).
```
puzzle-solver, line 28: 404, link not found

packages/ragbits-core/src/ragbits/core/llms/openai.py
```diff
@@ -0,0 +1,510 @@
...
+
+    HAS_TIKTOKEN = True
+except ImportError:
+    HAS_TIKTOKEN = False
+
+
+class OpenAILLMOptions(LLMOptions):
+    """
+    Dataclass that represents all available LLM call options for the OpenAI client.
+    Each of them is described in the [OpenAI API documentation](https://platform.openai.com/docs/api-reference/chat).
```
puzzle-solver, line 46: 404 not found. I think openai made their documentation a little bit more confusing recently.

packages/ragbits-core/src/ragbits/core/llms/litellm.py
puzzle-solver, line 45: `max_tokens` here are declared again, even though base `LLMOptions` already has it. I know that it's not a scope of this PR, but other classes don't provide it, so we need some consistency. I think some set of parameters can be potentially moved to the base LLMOptions class (also `temperature` and `top_p`). Or we can ditch the `LLMOptions` class completely, since it doesn't really provide much value.

packages/ragbits-core/src/ragbits/core/llms/openai.py
```diff
@@ -0,0 +1,510 @@
...
+        """
+        Returns the model id.
+        """
+        return "openai:" + self.model_name
+
+    def get_estimated_cost(self, prompt_tokens: int, completion_tokens: int) -> float:  # noqa: PLR6301
+        """
+        Returns 0.0 — cost estimation is not bundled; use OpenAI usage dashboard for billing details.
+        """
+        return 0.0
```
puzzle-solver, line 117: I mean, this is not too great. We probably shouldn't return 0 for a cost just like that if we actually have this method.

puzzle-solver: This is also the case for Gemini and Anthropic

packages/ragbits-core/src/ragbits/core/llms/openai.py
```diff
@@ -0,0 +1,510 @@
...
+        Returns:
+            Approximate token count.
+        """
+        if HAS_TIKTOKEN:
+            try:
+                enc = tiktoken.encoding_for_model(self.model_name)
+                return sum(len(enc.encode(str(msg.get("content") or ""))) for msg in prompt.chat)
+            except KeyError:
+                pass
+        return sum(len(str(msg.get("content") or "")) for msg in prompt.chat)
```
puzzle-solver, line 135: Not a good fallback, we should probably add `tiktoken` to dependencies explicitly when the user wants to use `openai`, otherwise they will silently get completely wrong token counts (number of characters is probably a few times higher than number of tokens)

puzzle-solver (commented):
Overall looks good, but we need to make sure that this is the direction we want to go in. I would suggest leaving the LiteLLM dependency to not make a breaking change yet and independently add OpenAI/Anthropic/Google providers. Also, we need to make sure that the implementations of them are correct, including token counting and cost estimation, which right now is broken. 

Eventually, I think adding something like `RagbitsLLM` that routes between the three providers would be cool (similarly to LiteLLM).  LiteLLM would be a fallback to other models if needed.

PR #977, opened 2026-04-24
feat: conversation sharing

JakPing:
feat(chat): add conversation sharing with SharePersistence and UI flow
    
    Introduce a SharePersistence layer, share/unshare API endpoints, and a SQL-backed
    implementation, and wire the React UI (ShareButton, ChatHistory, HistoryStore,
    ConversationGuard) through a new shares API client so conversations can be
    generated, fetched, and revoked via shareable links. Adds unit tests for share
    persistence and SQL conversation handling.


packages/ragbits-chat/src/ragbits/chat/providers/model_provider.py
```diff
@@ -198,6 +209,10 @@ def get_categories(self) -> dict[str, list[str]]:
                 "ConfigResponse",
                 "LoginResponse",
                 "OAuth2AuthorizeResponse",
+                "ConversationShareResponse",
+                "ConversationMeta",
+                "ConversationInteractionData",
+                "ConversationDetail",
```
jakubduda-dsai, line 215: Are all of these responses for sure? Or maybe some of them should be in core_data array?

typescript/ui/src/plugins/ChatHistoryPlugin/routes/ConversationGuard.tsx
```diff
@@ -1,41 +1,100 @@
...
+      "[Guard] effect fire " +
+        JSON.stringify({
+          conversationId,
+          isKnown,
+          isTemp,
+          isServerOnly,
+          loading,
+          conversations,
+        }),
+    );
```
jakubduda-dsai: console.log to remove

typescript/ui/src/core/types/history.ts
```diff
@@ -61,7 +67,10 @@ export interface HistoryStore {
   actions: {
     newConversation: () => string;
     selectConversation: (conversationId: string) => void;
-    deleteConversation: (conversationId: string) => void;
+    deleteConversation: (
+      conversationId: string,
+      ragbitsClient?: RagbitsClient,
```
jakubduda-dsai, line 72: shouldnt this type be as required?

typescript/ui/src/plugins/ChatHistoryPlugin/routes/ConversationGuard.tsx
```diff
@@ -1,41 +1,100 @@
...
+    const fallbackToHomeConversation = () => {
+      const newestConversation = conversations.at(-1);
+      console.log("[Guard] fallback " + JSON.stringify({ newestConversation }));
+      const target = newestConversation ?? newConversation();
+      console.log("[Guard] fallback target=" + target);
```
jakubduda-dsai: console.logs

typescript/ui/src/plugins/ChatHistoryPlugin/routes/ConversationGuard.tsx
jakubduda-dsai: log

packages/ragbits-chat/src/ragbits/chat/persistence/sql.py
```diff
@@ -282,6 +291,105 @@ async def get_conversation_interactions(self, conversation_id: str) -> list[dict
...
+                    "id": conv.id,
+                    "user_id": conv.user_id,
+                    "created_at": conv.created_at,
+                }
+                for conv in conversations
+            ]
+
+    async def get_conversation_summaries(self, conversation_ids: list[str]) -> dict[str, str]:
+        """
+        Get a display summary for each conversation, derived from the first user message.
```
puzzle-solver: nitpick: summaries don't have to be derived from the first user message. It may be done later in the conversation, if at any point `ChatInterface` yields a `ConversationSummaryResponse`

packages/ragbits-chat/src/ragbits/chat/persistence/sql.py
```diff
@@ -282,6 +291,105 @@ async def get_conversation_interactions(self, conversation_id: str) -> list[dict
...
+        Returns:
+            A list of conversation dictionaries.
+        """
+        await self._init_db()
+
+        async with AsyncSession(self.sqlalchemy_engine) as session:
+            result = await session.execute(
+                sqlalchemy.select(self.Conversation)
+                .filter_by(user_id=user_id)
+                .order_by(self.Conversation.created_at.desc())
```
puzzle-solver: How about ordering by `updated_at`? ChatGPT does that by default from what I see.

packages/ragbits-chat/src/ragbits/chat/persistence/sql.py
```diff
@@ -282,6 +291,105 @@ async def get_conversation_interactions(self, conversation_id: str) -> list[dict
...
+            for cid in conversation_ids:
+                result = await session.execute(
+                    sqlalchemy.select(self.ChatInteraction.message)
+                    .filter_by(conversation_id=cid)
+                    .order_by(self.ChatInteraction.id.asc())
+                    .limit(1)
+                )
+                first_msg = result.scalar_one_or_none()
+                if first_msg:
+                    summaries[cid] = first_msg[:max_len] + ("…" if len(first_msg) > max_len else "")
```
puzzle-solver, line 411: Ideally, the conversation summary should be taken directly from the `conversations` table rather than creating them by hand. However, `conversations` do not have a `summary` field. Thus, I think there are two approaches:
* add a `summary` field to the `conversations` table and update it each time we encounter a `conversation_summary` type of response in `extra responses` during `save_interaction`. This is nice, because then each conversation row has a summary 
* change just this method to iterate over all interactions for a given conversation and (similarly to the previous point), iterate over `extra_responses` to find a `conversation_summary`.

The second idea is probably easier to implement (just this method), but needs some unnecessary iteration every time we request `GET /api/conversations`

packages/ragbits-chat/src/ragbits/chat/api_routes/share.py
```diff
@@ -0,0 +1,168 @@
...
+    ConversationMeta,
+    ConversationShareResponse,
+    ShareConversationRequest,
+)
+from ragbits.chat.persistence.base import HistoryPersistenceStrategy, SharePersistenceStrategy
+
+RequireUser = Callable[[Request], Awaitable[User]]
+
+
+def build_share_router(  # noqa: PLR0915
```
puzzle-solver: I would suggest to name the whole file `conversations.py` and this function `build_conversations_router`, because IMO having the `/api/conversations` endpoint is a much more important functionality - allows us to have a backend conversation persistance. The sharing functionality is a nice addition. 

Also, I would also like to be able to use this router even when I don't necessarily have a `share_persistence` set up, just `history_persistence`. So I think some untangling here would be needed, but seems very doable with the current setup.

packages/ragbits-core/src/ragbits/core/storage/sql_store/__init__.py
```diff
@@ -0,0 +1,38 @@
+"""SQL store abstractions.
```
puzzle-solver, line 1: Are we planning to use this module? Right now, a persistence of chat history is done with a sqlalchemy ORM, which I quite like. With this approach, we would need to operate on untyped dicts.

Not strongly against it, but I would be interested to know what are we going to do with this code.

packages/ragbits-chat/src/ragbits/chat/api.py
```diff
@@ -301,12 +325,47 @@ async def theme() -> PlainTextResponse:
...
 
+    def _setup_share_routes(self) -> None:
+        """Register routes for conversation sharing.
+
+        Requires ``auth_backend``, ``share_persistence``, and a
+        ``history_persistence`` that supports ownership tracking (i.e. overrides
+        :meth:`HistoryPersistenceStrategy.list_conversations` to return the
+        user's conversations).
+        """
+        share = self.share_persistence
```
puzzle-solver: Let's just name it `share_persistence` for consistency with `history_persistence` and clarity

packages/ragbits-chat/src/ragbits/chat/api_routes/share.py
```diff
@@ -0,0 +1,168 @@
...
+    history: HistoryPersistenceStrategy,
+    share: SharePersistenceStrategy,
```
puzzle-solver: I would suggest naming it `history_persistence` and `share_persistence` for consistency (but not a must have)

packages/ragbits-chat/src/ragbits/chat/persistence/share.py
```diff
@@ -0,0 +1,363 @@
...
+                    "config": {"db_path": ":memory:"},
+                }
+            }
+        """
+        connection_options = ObjectConstructionConfig.model_validate(config["connection"])
+        config["connection"] = DatabaseConnection.subclass_from_config(connection_options)
+        return cls(**config)
+
+
+# Backwards-compatible alias kept while downstream code migrates to the
```
puzzle-solver: Should we delete it?

packages/ragbits-chat/src/ragbits/chat/persistence/share.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def _bool_literal(self, value: bool) -> str:
+        """Render a SQL boolean literal (some SQLite versions lack TRUE/FALSE)."""
+        if self._is_postgres:
+            return "TRUE" if value else "FALSE"
+        return "1" if value else "0"
+
+    @staticmethod
+    def _normalize_row(row: dict[str, Any]) -> dict[str, Any]:
+        """Normalise a raw row so callers see consistent types across dialects."""
```
puzzle-solver, line 138: nitpick / observation: I think with sqlalchemy (which we use for `SQLHistoryPersistence`) we wouldn't have to do these things. I think it would be good to use the same convention here, but this is certainly not a requirement.

puzzle-solver (commented):
Overall, I think that this PR looks very good and we definitely need to include that in a new release. I think many projects may benefit from having a backend conversation history persistence, since so far it had to be manually implemented.

Left some nit picks in the comments, but the main thing I would suggest is a separate set up of conversation endpoints for storing conversation and for sharing them. From my experience with client projects, we may often want to have server-side history persistence, without thinking about setting up conversation sharing.

packages/ragbits-chat/src/ragbits/chat/providers/model_provider.py
JakPing: Based on the current model semantics, those four are imo correctly placed under responses, not core_data.

typescript/ui/src/core/types/history.ts
JakPing: deleteConversation is designed to always remove locally, and only hit the API when a client is available, so i would leave it as is.

puzzle-solver (approved):

packages/ragbits-chat/src/ragbits/chat/api_routes/conversations.py
```diff
@@ -0,0 +1,186 @@
...
+        @router.delete("/api/shared/{conversation_id}", status_code=status.HTTP_204_NO_CONTENT)
+        async def dismiss_share(conversation_id: str, user: User = authed_user) -> Response:
+            if not await share_persistence.hide_share(conversation_id, recipient_identifiers(user)):
+                raise HTTPException(status_code=404, detail="Shared conversation not found.")
+            return Response(status_code=status.HTTP_204_NO_CONTENT)
+
+    return router
+
+
+def build_share_router(
```
puzzle-solver, line 176: I would delete this one, since this is a new PR and no existing code uses "build_share_router"

puzzle-solver (approved):

Issue #913, opened 2026-01-23
feat: support creating tools from functions with model parameters


puzzle-solver:
### Feature description

I would want to be able to use tools with defined by functions with parameters typed as pydantic models.

### Motivation

Example use case: creating a cart with a list of items

```
class Cart(BaseModel):
    products: list[Product]
    
class Product(BaseModel):
    product_id: str
    quantity: int = Field(default=1, gt=0)
```

Then, I would want to define my tool like this:

```
def create_cart(cart: Cart) -> bool:
    """Creates a cart with items that a user can order"""
    if validate(cart):
        return "success"
    else:
        return "cart creation failed"
```

The agent definition would be just `Agent(llm=llm, prompt=prompt, tools=[create_cart])`. 

---

Right now, specifying the tools this way does not capture the schema of a Cart  and is not reflected in JSON schema that is created and send to the API.

The workaround for now is to use a custom tool:
```
def on_create_cart_tool_call(products: list):
    cart = Cart(products=products)
    return create_cart(cart)

Tool(
        name="create_cart",
        description="Creates a cart with items that a user can order",
        parameters=Cart.model_json_schema(),
        on_tool_call=on_create_cart_tool_call,
    )
```

However, this is much less clean.

PR #978, opened 2026-05-05
feat: support pydantic models in tool arguments

puzzle-solver:
Solves https://github.com/deepsense-ai/ragbits/issues/913

Now, we can pass pydantic models as tool arguments. The flow is the following: 
* define a function with pydantic model as argument -
* pass it to agent 
* LLM receives the appropriate JSON schema 
* LLM sends the data in JSON format 
* tool converts that JSON into the pydantic model, so that the callable can run

### DIFF
--- a/packages/ragbits-agents/src/ragbits/agents/tool.py
+++ b/packages/ragbits-agents/src/ragbits/agents/tool.py
@@ -3,7 +3,7 @@ from contextlib import suppress
 from dataclasses import dataclass
 from typing import TYPE_CHECKING, Any, Literal, cast
 
-from pydantic import BaseModel
+from pydantic import BaseModel, validate_call
 from typing_extensions import Self
 
 from ragbits.core.llms.base import LLMClientOptionsT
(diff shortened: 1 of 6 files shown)


packages/ragbits-agents/src/ragbits/agents/tool.py
```diff
@@ -93,7 +93,7 @@ def from_callable(cls, callable: Callable) -> Self:
             name=schema["function"]["name"],
             description=schema["function"]["description"],
             parameters=schema["function"]["parameters"],
-            on_tool_call=callable,
+            on_tool_call=validate_call(callable),
```
puzzle-solver, line 102: This piece of code is the only thing I am a little bit concerned about. It may happen, that the LLM even with the specified JSON schema returns a dictionary that doesn't match that schema. If the user did the validation themselves, they could catch the error and return the information about it to the LLM. With this, we are just raising an error.  But I think that if this becomes a concern, a user may always create a `Tool` object directly with their own callable.

JakPing (approved):

PR #981, opened 2026-05-18
feat: improved file upload frontend and add attachment validation

puzzle-solver:
Added additional frontend+backend validation of uploads (configurable from the backend). Improved UI showing the attached files. 

image


typescript/ui/src/core/utils/fileHandlers.ts
```diff
@@ -0,0 +1,16 @@
...
+  return () => {
+    if (handler === h) handler = null;
+  };
+}
+
+export function dispatchDroppedFiles(files: File[]): boolean {
+  if (!handler || files.length === 0) return false;
+  handler(files);
+  return true;
+}
```
dazy-ds, line 16: I assume that this was added as a way of allowing plugins to register for file events, right?

If so, this is kind of a big architecture change/new feature and I think it should be discussed.

The implementation of this seems fine, as it somewhat mimics how we handle slots, etc. but I would like to know rationale behind that implementation.

On the side note, if we decide to keep it, I would move it to a directory called `registires` or `events` or something similar. That way we can communicate that those methods are used to register some handlers of global events emitted by the UI (just like in an even driven application, e.g. CMSes have similar mechanisms).

typescript/ui/src/core/utils/messageContext.ts
```diff
@@ -0,0 +1,36 @@
...
+      userMessageExtra = { ...(userMessageExtra ?? {}), ...r.userMessageExtra };
+    }
+    if (r.files && r.files.length > 0) files.push(...r.files);
+  }
+  return {
+    context,
+    userMessageExtra,
+    files: files.length > 0 ? files : undefined,
+  };
+}
```
dazy-ds, line 36: Same as above.

typescript/ui/src/plugins/UploadPlugin/stores/attachmentsStore.ts
```diff
@@ -1,35 +1,131 @@
...
+        id: crypto.randomUUID(),
+        filename: file.name,
+        mimeType: file.type || "application/octet-stream",
+        blobUrl: URL.createObjectURL(file),
+        status: "failed",
+        error,
+        retryFile,
+      };
+      set((state) => ({ pending: [...state.pending, entry] }));
+    },
```
dazy-ds, line 88: This seems like it could be merged into one method. We could shift the setting of "status" to consumer of the store and expect objects of a specified structure as arguments (to enforce the required arguments).

E.g.
```typescript
interface FailedFile {
  file: File,
  status: "failed",
  error: string,
  retryFile?: File
}

interface UploadingFile {
  file: File,
  status: "uploading"
}
add(fileObject: UploadingFile | FailedFile) {}
```

I want to note that this is a pure cosmetic change coming from my preference :D

We can also keep this as is, but abstract the common `entry` creation process (it would give the id, extract filename, etc), and keep the case specific code in the existing methods.

typescript/ui/src/plugins/UploadPlugin/index.tsx
```diff
@@ -1,14 +1,53 @@
...
+        .consumeReady();
+      if (files.length === 0) return null;
+      return {
+        files,
+        userMessageExtra: { attachments: previews },
+      };
+    });
+    unregisterDropHandler = registerFileDropHandler((files) => {
+      files.forEach(processFile);
+    });
```
dazy-ds, line 44: Registering this way is okay (although I am not a fan of usage of global variables for such purposes).

We can also consider expanding the Plugin to support a list of callbacks to be returned from `onActivate` that would be automatically called when plugin is deactivated. It can simplify the plugin registration process while also keeping `onDeactivate` for some general stuff.

Either way I am fine with that approach, just my thought.

typescript/ui/src/ragbits/stores/HistoryStore/historyStore.ts
```diff
@@ -369,17 +369,14 @@ export const createHistoryStore = immer<HistoryStore>((set, get) => ({
...
+      if (files && files.length > 0) {
         const formData = new FormData();
         formData.append("request", JSON.stringify(chatRequest));
-        for (const file of pendingFiles) {
+        for (const file of files) {
           formData.append("files", file, file.name);
         }
         requestBody = formData;
-        attachmentsStore.getState().clear();
       }
```
dazy-ds, line 380: This is the part that got most of my attention in this PR.

Previous version negated whole purpose of having plugins, because it directly imported a file from a plugin into one of the core entities. The fixed version is better, but I am still not sure about splitting `files` and `userMessageExtra` objects.

One can also wonder why would we add a utility to add files into the core, other that for uploading them. I can't see other use case for files dropped into the chat, other than them being uploaded to be processed (or processed on the client). Having that in mind, we could consider adding whole `UploadPlugin` into the core.

To summarize I think this is fine with a small nit about naming and separating files from `userMessageExtra`, because for me, files are extra additions to user's message.

typescript/ui/src/plugins/UploadPlugin/validation.ts
```diff
@@ -0,0 +1,43 @@
...
+      return "Uploads are currently unavailable.";
+    case "too-many":
+      return `Too many attachments — limit is ${error.limit} per message.`;
+    case "too-large": {
+      const limitMb = Math.round(error.limitBytes / (1024 * 1024));
+      return `File too large — limit is ${limitMb} MB.`;
+    }
+    case "wrong-type":
+      return `File type not supported. Allowed: ${error.allowed.join(", ")}.`;
+  }
```
dazy-ds, line 42: nit: We can add exhaustive check to always make sure that we have all cases handled in case a new kind is added (https://stackoverflow.com/questions/39419170/how-do-i-check-that-a-switch-block-is-exhaustive-in-typescript)

dazy-ds (changes requested):
Overall looks good, I would like to have some additional context before approving. I just want to fully understand the rationale behind the implementation C:

Also left some food for thoughts :D

